PR #6, opened 2019-11-27
Initial Auth0 collector implementation

jeremey:
Implement an Auth0 log collector using the `paws-collector` library/framework.

### DIFF
--- a/cfn/auth0-collector.template
+++ b/cfn/auth0-collector.template
@@ -1,0 +1,809 @@
+{
+    "AWSTemplateFormatVersion": "2010-09-09",
+    "Description": "Alert Logic template for creating a Auth0 Logs collector",
+    "Parameters": {
+        "AlertlogicAccessKeyId": {
+            "Description": "Alert Logic Access Key Id obtained from AIMS",
+            "Type": "String",
+            "Default": "5c850016c9eb6f3b"
+        },
+        "AlertlogicSecretKey": {
+            "Description": "Alert Logic Secret Key returned from AIMS for the Access Key Id",
+            "Type": "String",
+            "NoEcho": true
+        },
+        "AlApiEndpoint": {
+            "Description": "Alert Logic API endpoint",
+            "Type": "String",
+            "Default": "api.global-services.global.alertlogic.com",
+            "AllowedValues": [
+                "api.global-services.global.alertlogic.com",
+                "api.global-integration.product.dev.alertlogic.com"
+            ]
+        },
+        "AlDataResidency": {
+            "Description": "Alert Logic Data Residency",
+            "Type": "String",
+            "Default": "default",
+            "AllowedValues": ["default"]
+        },
+        "PackagesBucketPrefix": {
+            "Description": "S3 bucket name prefix where collector packages are located.",
+            "Type": "String",
+            "Default": "rcs-alertlogic-collectors"
+        },
+        
+        "AlertlogicCustomerId": {
+            "Description": "Optional, Alert Logic customer ID which collected data should be reported for. If not set customer ID is derived from AIMs tokens",
+            "Type": "String",
+            "Default": ""
+        },
+        "CollectorId": {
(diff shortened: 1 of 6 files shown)


jeremey:
@kkuzmin no local or test directory here yet, have a couple questions about them.

collectors/auth0/index.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+const debug = require('debug') ('index');
+
+const Auth0Collector = require('./auth0_collector').Auth0Collector;
+
+exports.handler = function(event, context) {
+    debug('input event: ', event);
+    Auth0Collector.load().then(function(creds) {
+        var auth0c = new Auth0Collector(context, creds, 'auth0');
```
kkuzmin: no need to pass 3rd parameter here.

collectors/auth0/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "devDependencies": {
+    "jshint": "^2.9.5",
+    "mocha": "^6.2.2",
+    "mocha-jenkins-reporter": "^0.4.2",
+    "nyc": "^14.1.1",
+    "rewire": "^4.0.1",
+    "sinon": "^7.5.0"
+  },
+    "dependencies": {
+    "auth0": "^2.20.0",
```
kkuzmin: Let's remove `^`
We normally hard pin dependencies.

collectors/auth0/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "@alertlogic/al-collector-js": "1.3.4",
+    "@alertlogic/paws-collector": "1.0.0",
```
kkuzmin: please use 1.0.3

collectors/auth0/auth0_collector.js
```diff
@@ -0,0 +1,105 @@
...
+    
+    pawsGetLogs(state, callback) {
+        let collector = this;
+        const auth0Client = new ManagementClient({
+            domain: process.env.paws_domain,
+            clientId: process.env.paws_client_id,
+            clientSecret: process.env.paws_api_secret,
+            scope: 'read:logs'
+        });
+        let params = state.last_log_id ? {from: state.last_log_id} : {q: "date=[" + state.since + " TO *]"};
```
kkuzmin: Looks like `state.since` is not updated in `_getNextCollectionState()`. Is that ok?

cfn/auth0-collector.template
```diff
@@ -0,0 +1,809 @@
+{
+    "AWSTemplateFormatVersion": "2010-09-09",
+    "Description": "Alert Logic template for creating a Auth0 Logs collector",
+    "Parameters": {
+        "AlertlogicAccessKeyId": {
+            "Description": "Alert Logic Access Key Id obtained from AIMS",
+            "Type": "String",
+            "Default": "5c850016c9eb6f3b"
```
kkuzmin: please remove default. It was used for testing.
I'll remove it for okta as well.

collectors/auth0/index.js
jeremey: k, was following okta example.

collectors/auth0/auth0_collector.js
jeremey: yes, it is only used for the initial fetch, after that we always fetch based on a `last_log_id`.

collectors/auth0/index.js
kkuzmin: okta example still needs some clean up as well

collectors/auth0/index.js
```diff
@@ -0,0 +1,23 @@
...
+        auth0c.handleEvent(event, function(err) {
```
kkuzmin: this function callback is not required as well.

kkuzmin: you can simply use it like
```
auth0c.handleEvent(event);
```

PR #20, opened 2019-12-10
Okta/Auth0 update with the latest PAWS

kkuzmin:
### Solution Description
Depends on https://github.com/alertlogic/paws-collector/pull/19
Use latest PAWS with encrypted secrets.

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -26,9 +26,6 @@ const tsPaths = [
 
 
 class OktaCollector extends PawsCollector {
-    constructor(context, creds) {
-        super(context, creds, 'okta');
-    }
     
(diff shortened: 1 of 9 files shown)


collectors/okta/okta_collector.js
```diff
@@ -26,9 +26,6 @@ const tsPaths = [
 
 
 class OktaCollector extends PawsCollector {
-    constructor(context, creds) {
-        super(context, creds, 'okta');
-    }
```
kkuzmin: Removed, because default constructor looks exactly same.

kkuzmin:
@JDragovichAlertLogic the okta temaplte is not migrated yet. I'd like to have a smoke test with the `paws-collector.template` first and then send a follow-up PR.

PR #22, opened 2019-12-11
Initial CLI driver for testing collection itself

jeremey:
### Problem Description
Unnecessary effort is required to simply test that the collection implementation in a paws collector works.

### Solution Description
Provide a simple command-line driver that will exercise the collection aspect of a collector independent of any Alert Logic identity, registration, transport, etc.

This is not a substitute for automated tests but it will confirm that the collector implementation actually collects logs.

I suggest we integrate this into the collector template/generator also (separate PR).


jeremey:
Closing this one to rebase/adapt off #20

PR #28, opened 2019-12-20
Add test coverage for o365

JDragovichAlertLogic:
Add test coverage for o365


collectors/o365/lib/o365_mgmnt/o365management.js
```diff
@@ -262,52 +131,6 @@ class O365Management extends msRestAzure.AzureServiceClient {
...
-                    httpRequest.headers.set(headerName, options.customHeaders[headerName]);
-                }
-            }
-        }
-        httpRequest.headers.set('Content-Type', 'application/json; charset=utf-8');
-        httpRequest.body = null;
-        // Create handler
-        const handler = this.requestHandler(httpRequest);
-        // Send Request
-        return client.sendRequest(httpRequest).then(handler);
```
OHaganA: Is all of this from the old SDK and that's why it's being nuked?

JDragovichAlertLogic: yeah. It was from the old collector and not being used. Didnt feel like it was worth writing tests for it!

PR #32, opened 2020-01-01
Google Cloud Stackdriver POC

JDragovichAlertLogic:
Here is a POC for a PAWS based google stackdriver collector. This needs to some more polishing and testing in AWS. It successfully collects logs and sends them to Ingest Locally. This is an experiment to learn a bit about Google Cloud and maybe get a start on another PAWS type collector. *This is a POC not ready for Production*

Caveats
- We needs to understand if we need to decode the actual protobuff payload in the log entry.
- We need to see if we should/could get logs from a Google Cloud Organization. Right now this only gets logs from a collection of Projects (the equivalent of an AWS VPC or Azure RG as far as I can tell), but Organizations sit above projects.
- As always, this needs some unit tests to be finished.


collectors/googlestackdriver/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Generate Google API key file
+1. In the Google Cloud Console, go to the Create service account key page.
+2. From the Service account list, select New service account.
+3. In the Service account name field, enter a name.
+4. From the Role list, select Project > Owner.
+
+Note: The Role field authorizes your service account to access resources. You can view and change this field later by using the Cloud Console. If you are developing a production app, specify more granular permissions than Project > Owner. For more information, see granting roles to service accounts.
+
+5. Click Create. A JSON file that contains your key downloads to your computer.
```
OHaganA: Nothing actually wrong here, but wondering if this could be automated? Seeing the word "Service" in reference to a collector reminded me of Azure and filled me with dread

collectors/googlestackdriver/README.md
```diff
@@ -0,0 +1,101 @@
...
+  3. Fill in environment variables in `env.json` (including encrypted AIMS secret key) and invoke locally:
+
+```
+cp ./local/env.json.tmpl ./local/env.json
+vi ./local/env.json
+make test
+make sam-local
+```
+  4. Please see `local/event.json` for the event payload used for local invocation.
+Please write your readme here
```
OHaganA: Placeholder line that needs removed

collectors/googlestackdriver/test/mock.js
```diff
@@ -0,0 +1,31 @@
...
+process.env.googlestackdriver_token = 'googlestackdriver-token';
+process.env.collector_id = 'collector-id';
+
+const AIMS_TEST_CREDS = {
+    access_key_id: 'test-access-key-id',
+    secret_key: 'test-secret-key'
+};
+
+const LOG_EVENT = {
+    // Here is where you fill in you mock log event
```
OHaganA: No mock log event required here?

JDragovichAlertLogic: Like i said in the PR description, this is a POC and does not have tests yet.

collectors/googlestackdriver/README.md
JDragovichAlertLogic: It is similar to Azure, but mich more straightforward. Users would still need to log into their consoles, but once its created, its simply a matter of downloading the JSON and using it in the collector.

cfn/googlestackdriver_collector.template
```diff
@@ -0,0 +1,858 @@
...
+            "Type": "Number",
+            "Default": 60
+        },
+        "GoogleProjectIds":{
+            "Description": "JSON array of porject IDs from which to draw logs",
+            "Type": "string",
+            "Default": "[]"
+        },
+        "PawsSecret": {
+            "Description": "Paste the contents of your JSON credentail file here.",
```
OHaganA: Just caught this typo `credentail` whilst looking at this PR, but as these templates are boilerplate this typo is most likely in all paws collector templates.

collectors/googlestackdriver/test/mock.js
OHaganA: Oh yep, of course, tests are wip, sorry

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,120 @@
...
+
+
+        console.info(`GSTA000001 Collecting data from ${state.since} till ${state.until}`);
+
+        // TODO: figure out a better way to format this. I'm pretty sure that it needs the newlines in it.
+        const filter = `timestamp >= "${state.since}"
+timestamp < "${state.until}"`;
+        const formattedResourceNames = JSON.parse(process.env.google_project_ids).map(e => `projects/${e}`);
+
+        // TODO: check out how the "autopagination" functionality works on this.
```
kkuzmin: there is a `nextPageToken` field which need to be processed

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,120 @@
...
+        client.listLogEntries({filter:filter, resourceNames: formattedResourceNames})
+            .then(responses => {
+                const resources = responses[0];
```
kkuzmin: why only first element instead of entire array?

collectors/googlestackdriver/package-lock.json
```diff
@@ -0,0 +1,4203 @@
+{
```
kkuzmin: this file is not needed.

collectors/googlestackdriver/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "mocha": "^6.2.2",
+    "mocha-jenkins-reporter": "^0.4.2",
+    "nyc": "^14.1.1",
+    "rewire": "^4.0.1",
+    "sinon": "^7.5.0"
+  },
+  "dependencies": {
+    "@alertlogic/al-collector-js": "1.4.2",
+    "@alertlogic/paws-collector": "1.0.8",
+    "@google-cloud/logging": "^6.0.0",
```
kkuzmin: please pin the version

collectors/googlestackdriver/test/mock.js
kkuzmin: a sample event will be greatly appreciated. Event without tests =)

collectors/googlestackdriver/collector.js
kkuzmin: Or pass `{autoPaginate: true}` in the options, like described here https://cloud.google.com/nodejs/docs/reference/logging/2.0.x/v2.LoggingServiceV2Client#listLogEntries

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,120 @@
...
+            credentials: JSON.parse(this._pawsCreds.secret)
+        });
+
+
+        console.info(`GSTA000001 Collecting data from ${state.since} till ${state.until}`);
+
+        // TODO: figure out a better way to format this. I'm pretty sure that it needs the newlines in it.
+        const filter = `timestamp >= "${state.since}"
+timestamp < "${state.until}"`;
+        const formattedResourceNames = JSON.parse(process.env.google_project_ids).map(e => `projects/${e}`);
```
kkuzmin: Is it possible to specify `all` projects?

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,120 @@
...
+        return  {
+             since: nextSinceTs,
+             until: nextUntilMoment.toISOString(),
+             poll_interval_sec: nextPollInterval
+        };
+    }
+    
+    // TODO: probably need to actually decode hte protobuf payload on these logs
+    pawsFormatLog(msg) {
+        const ts = msg.timestamp ? msg.timestamp : {seconds: Date.now() / 1000};
```
kkuzmin: Is there any reason to not use `tsPaths`, like https://github.com/alertlogic/paws-collector/blob/master/collectors/okta/okta_collector.js#L90?

kkuzmin: This is where a sample event would help.

collectors/googlestackdriver/collector.js
JDragovichAlertLogic: So My understanding is that autopagination is on by default. But that is based on this example i used:

https://googleapis.dev/nodejs/logging/latest/v2.LoggingServiceV2Client.html#listLogEntries-examples

There are two methods shown, the first doesn't set auto pagination, which is the one i used.

Though I admit I cannot find confirmation in the api docs that this option is true by default

collectors/googlestackdriver/collector.js
JDragovichAlertLogic: see example. 

https://googleapis.dev/nodejs/logging/latest/v2.LoggingServiceV2Client.html#listLogEntries-examples

It seems that the actual content is an array that is the first element of that array.

collectors/googlestackdriver/collector.js
kkuzmin: I think it is safer to explicitly set it to `true`.

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,120 @@
...
+        };
+        return callback(null, initialState, 1);
+    }
+    
+    pawsGetLogs(state, callback) {
+        let collector = this;
+
+        // Start API client
+        const client = new logging.v2.LoggingServiceV2Client({
+            credentials: JSON.parse(this._pawsCreds.secret)
```
kkuzmin: You can use `collector.secret` here.

collectors/googlestackdriver/package-lock.json
JDragovichAlertLogic: I still think that its better to include these lock files. We use similar ones in our erlang services. However we do hard pin libraries, so it might be less important.

collectors/googlestackdriver/collector.js
JDragovichAlertLogic: I don't think that's possible to per this api doc. https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/list

Thinking about it, it wouldnt make sense. like i don't think you can collector logs form every vpc in AWS or every rg in Azure.

You can specify an organization, which can sit above several projects. I think that will collect logs from several projects, but I have not validated that idea.

kkuzmin: In Azure it is possible to get activity logs for all RGs. Should be something similar for collection all logs here.
Maybe here we can use `projects/*`.
Actually I think we can replace `google_project_ids` with `google_resource_names` and simply pass this parameter into `client.listLogEntries()` as is. Thoughts?

JDragovichAlertLogic: So in my own tests with teh api explorer, it does not accept `projects/*`

But yeah we could make that parameter resource names, but then they would need to be in the format `<resource-type>/<name>` in the param. I just restricted to project because that was 1. what i could test an 2. seemed to be the main 'unit of organization' for google cloud.

kkuzmin: I'd still stick to `google_resource_names` because logs can be retrieved for multiple resource types.
> "projects/[PROJECT_ID]"
"organizations/[ORGANIZATION_ID]"
"billingAccounts/[BILLING_ACCOUNT_ID]"
"folders/[FOLDER_ID]"

JDragovichAlertLogic: Thats fair. I'll change it then.

collectors/googlestackdriver/collector.js
JDragovichAlertLogic: its because the TS that comesm back is not an iso timestamp, but is already returned in epoch time.

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,123 @@
...
+const parse = require('@alertlogic/al-collector-js').Parse;
+const logging = require('@google-cloud/logging');
+
+
+const typeIdPaths = [
+    {path: ['protoPayload', 'type_url']}
+];
+
+class GooglestackdriverCollector extends PawsCollector {
+    constructor(context, creds) {
```
kkuzmin: please remove. Default constructor can be used

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,123 @@
...
+            progName: 'GooglestackdriverCollector',
+            message: JSON.stringify(msg),
+            messageType: 'json/googlestackdriver'
+        };
+        
+        if (typeId !== null && typeId !== undefined) {
+            formattedMsg.messageTypeId = `${typeId}`;
+        }
+        if (ts.nanos) {
+            formattedMsg.messageTsUs = ts.usec;
```
kkuzmin: should be `ts.nanos`?

collectors/googlestackdriver/test/mock.js
```diff
@@ -0,0 +1,48 @@
...
+process.env.googlestackdriver_endpoint = 'https://test.alertlogic.com/';
+process.env.googlestackdriver_token = 'googlestackdriver-token';
+process.env.collector_id = 'collector-id';
+
+const AIMS_TEST_CREDS = {
+    access_key_id: 'test-access-key-id',
+    secret_key: 'test-secret-key'
+};
+
+const LOG_EVENT = { labels: {},
```
kkuzmin: PLease add samples for JSON and Text payload types.

cfn/googlestackdriver_collector.template
```diff
@@ -0,0 +1,94 @@
...
+          "AlApiEndpoint" : { "Ref":"AlApiEndpoint" },
+          "AlDataResidency" : { "Ref":"AlDataResidency" },
+          "PackagesBucketPrefix" : { "Ref":"PackagesBucketPrefix" },
+          "PawsCollectorTypeName" : "googlestackdriver",
+          "AlertlogicCustomerId" : { "Ref":"AlertlogicCustomerId" },
+          "CollectorId" : { "Ref":"CollectorId" },
+          "PollingInterval" : 60,
+          "PawsEndpoint" : { "Ref":"Auth0Domain" },
+          "PawsAuthType" : "oauth2",
+          "PawsClientId" : { "Ref":"Auth0ClientId" },
```
kkuzmin: Auth0Domain, Auth0ClientId? Lease check for copy-paste

JDragovichAlertLogic: (disappear)

cfn/googlestackdriver_collector.template
```diff
@@ -0,0 +1,94 @@
...
+            "NoEcho": true
+        },
+        "GoogleResourceIds": {
+            "Description": "JSON list of google resources to poll logs from. In the format <resourceType>/<resourceID>",
+            "Type": "String"
+        },
+        "CollectionStartTs": {
+            "Description": "Timestamp when log collection starts. For example, 2020-01-13T16:00:00Z",
+            "Type": "String",
+            "AllowedPattern" : "^\\d{4}(-\\d{2}){2}T(\\d{2}:){2}\\d{2}Z$"
```
kkuzmin: please use updated pattern, https://github.com/alertlogic/paws-collector/pull/45/files#diff-71ea0d086916ed2ea4f6201343d529bdR63

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,149 @@
...
+const moment = require('moment');
+const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
+const parse = require('@alertlogic/al-collector-js').Parse;
+const logging = require('@google-cloud/logging');
+
+
+const typeIdPaths = [
+    // TODO: determine if there are any solid paths in JSON and Text payloads. I think these payloads are arbitrary
+    {path: ['protoPayload', 'type_url']}
+];
```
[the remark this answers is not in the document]
kkuzmin: and I think we always want to have payload as the last path, i.e. default `typeId` value if we can't find more reasonable one
 ```
{path: ['payload']}
```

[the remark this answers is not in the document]
Comment (JDragovichAlertLogic) [reply collectors/googlestackdriver/collector.js]: Sure thing. I can add these paths in.

I think that we have a larger problem with Stackdriver in that You can feed a lot of different logs in there, even logs from other clouds (you can feed Cloudtrail log in there for example).

I think thee are some sensible defaults but I do really think that the `jsonPayload` and the `textPayload` cannot be expected to conform to any sort of schema that we can derive types and time stamps from. Stackdriver is also Google's solution for application logs. its really just a bucket that you can dump any old logs into. Thats what I understand about it at least.

So we can make some sensible guesses, but I think the data coming from this collector is going to be fairly irregular. and we dont really have a good way fo knowing what those logs will look like until we see it in the wild I think.

now we could tyr and filter just the GCP audit logs, but that has problmes of its own I think

File: collectors/googlestackdriver/collector.js [hunk 18]
```diff
@@ -92,6 +104,20 @@ timestamp < "${state.until}"`;
             })
             .catch(err => {
                 console.error(`GSTA000003 err in collection ${err}`);
+                
+                // Stackdriver Logging api has some rate limits that we might run into.
+                // If we run inot a rate limit error, instead of returning the error,
+                // we return the state back to the queue with an additional second added, up to 10
+                // https://cloud.google.com/logging/quotas
+                if(err.match(/RESOURCE_EXHAUSTED/)){
```
kkuzmin: maybe we get some error code here we can use instead heavy `match()`?

collectors/googlestackdriver/collector.js
```diff
@@ -16,17 +16,29 @@ const logging = require('@google-cloud/logging');
 
 
 const typeIdPaths = [
-    // TODO: determine if there are any solid paths in JSON and Text payloads. I think these payloads are arbitrary
-    {path: ['protoPayload', 'type_url']}
+    {path: ['jsonPayload', 'fileds', 'event_type', 'stringValue']},
```
kkuzmin: typo, `fields`

collectors/googlestackdriver/collector.js
JDragovichAlertLogic: Good point. Yeah it looks like there are actually error codes in these errors. i think they show up nicely as strings becuase google was fancy with the way they are stringified. 

This mathc doesnt seem to work anyway. looks like these are more complex than just stirngs.

collectors/googlestackdriver/collector.js
```diff
@@ -0,0 +1,183 @@
...
+                return callback(null, logs, newState, newState.poll_interval_sec);
+            })
+            .catch(err => {
+                console.error(`GSTA000003 err in collection ${err}`);
+                
+                // Stackdriver Logging api has some rate limits that we might run into.
+                // If we run inot a rate limit error, instead of returning the error,
+                // we return the state back to the queue with an additional second added, up to 10
+                // https://cloud.google.com/logging/quotas
+                if(err.code === 8){
```
kkuzmin: please mention that 8 is a limit exceed error code. Or better use it as a global var, like
```
const API_THROTTLING_ERROR = 8;
```

PR #34, opened 2020-01-06
General Cleanup

JDragovichAlertLogic:
## Problem description
some stuff needed cleaned up

## Solution
♫ Scrub scrub here
Scrub scrub there 
change that function clause
it's time to clean the codebase up
in the merry old land of paws ♫


JDragovichAlertLogic:
More changes are on the way. need to clean up some linter stuff.

PR #39, opened 2020-01-08
Fix checkin permissions

kkuzmin:
### Problem Description
`cloudformation:DescribeStacks` only supports `*` resource

### Solution Description
Fix `DescribeStacks` permission


kkuzmin:
@JDragovichAlertLogic it is there, see `collector.template.template`

JDragovichAlertLogic:
So it is. My mistake. +1

PR #40, opened 2020-01-08
Do some fixes for o365 collector

JDragovichAlertLogic:
### Problem Description
some fixes for o365 in hopes of it being read for SKO. will continue tomorrow.

### Solution Description
added:
- fix for start times that aremore recent for 24 hours
- some pagination for list content

### DIFF
--- a/collectors/o365/lib/o365_mgmnt/o365management.js
+++ b/collectors/o365/lib/o365_mgmnt/o365management.js
@@ -169,7 +169,7 @@ class O365Management extends msRestAzure.AzureServiceClient {
         // Request Handler
         const handler = this.requestHandler(httpRequest);
         // Send Request
-        return client.sendRequest(httpRequest).then(handler)
+        return client.sendRequest(httpRequest).then(handler);
     }
 }
 
--- a/collectors/o365/lib/o365_mgmnt/o365management.js
+++ b/collectors/o365/lib/o365_mgmnt/o365management.js
@@ -67,6 +67,25 @@ class O365Management extends msRestAzure.AzureServiceClient {
         }
     }
 
+    // standard request handler
+    requestHandler(httpRequest){
+        return (res) => {
+            let {parsedBody, headers, bodyAsText, status} = res;
+            if(!parsedBody){
+                parsedBody = JSON.parse(bodyAsText);
+            }
+            if (status !== 200) {
+                let error = new Error(parsedBody);
+                error.statusCode = status;
+                error.request = msRest.stripRequest(httpRequest);
+                error.response = msRest.stripResponse(res);
+                throw error
+            }
+            const nextPageUri = headers.get('NextPageUri');
+            return { parsedBody, nextPageUri };
+        }
+    }
+    
     subscriptionsContent(contentType, startTs, endTs, options) {
         let client = this;
         // Construct URL
@@ -107,29 +126,13 @@ class O365Management extends msRestAzure.AzureServiceClient {
         }
         httpRequest.headers.set('Content-Type', 'application/json; charset=utf-8');
         httpRequest.body = null;
+        // Request Handler
+        const handler = this.requestHandler(httpRequest);
         // Send Request
-        return client.sendRequest(httpRequest).then((res) => {
-            let {parsedBody, headers, bodyAsText, status} = res;
-            if(!parsedBody){
-                parsedBody = JSON.parse(bodyAsText);
-            }
-            if (status !== 200) {
-                let error = new Error(parsedBody);
-                error.statusCode = status;
-                error.request = msRest.stripRequest(httpRequest);
-                error.response = msRest.stripResponse(res);
-                throw error
-            }
-            const nextPage = headers.get('NextPageUri');
-            if(nextPage){
-                console.log('subsequent page detected calling: ', nextPage);
-                return this.getNextSubscriptionsContentPage(parsedHeaders.NextPageUri, parsedBody, options);
-            }
-            return parsedBody;
-        });
+        return client.sendRequest(httpRequest).then(handler);
     }
     
-    getNextSubscriptionsContentPage(nextPageUrl, contentAcc, options) {
+    getNextSubscriptionsContentPage(nextPageUrl, options) {
         let client = this;
     
         const httpOptions = {};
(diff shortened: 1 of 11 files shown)


collectors/o365/lib/o365_mgmnt/o365management.js
```diff
@@ -67,25 +67,7 @@ class O365Management extends msRestAzure.AzureServiceClient {
         }
     }
 
-    // standard request handler
-    requestHandler(httpRequest){
```
kkuzmin: I suggest we keep existing functions as is and add a new one, like, `getNextPageUrl(nextPageUrl, options)` helper function which simply retrieves the next page. This approach gives the called full control over pagination and then we can have the actual page listing inside `o365_collector.js`. We may want to have some pagination configuration (for example, how many pages we'd like to process before sending data to ingest) which we'll be able to add via template params or function configuration.

collectors/o365/lib/o365_mgmnt/o365management.js
```diff
@@ -125,12 +126,42 @@ class O365Management extends msRestAzure.AzureServiceClient {
...
         httpRequest.headers.set('Content-Type', 'application/json; charset=utf-8');
         httpRequest.body = null;
-        // Create handler
+        // Request Handler
         const handler = this.requestHandler(httpRequest);
         // Send Request
         return client.sendRequest(httpRequest).then(handler);
     }
     
+    getNextSubscriptionsContentPage(nextPageUrl, options) {
```
kkuzmin: I'm sure this function can be used for any paged content.
I suggest we call it `getNextPageUrl`.
BTW, can't we use `getContent()` function for requesting nextPagetUrl?

JDragovichAlertLogic: It can, and that was the original aim. I'll rename it.

and you are correct. We could just use `getContent` to retrieve the next page. All it's doing right now is just getting the body at the end.

collectors/o365/o365_collector.js
```diff
@@ -49,13 +46,21 @@ class O365Collector extends PawsCollector {
             endTs = moment(startTs).add(24, 'hours').toISOString();
```
kkuzmin: please check `console.log` message above and change it to `console.info/warn` with some message id.

collectors/o365/o365_collector.js
```diff
@@ -73,28 +78,63 @@ class O365Collector extends PawsCollector {
...
-            const contentPromises = aggregateResult.map(({contentUri}) => m_o365mgmnt.getContent(contentUri));
-            return Promise.all(contentPromises);
-        // Call all of the content urls and agregate the resutls
-        }).then(content => {
-            const flattenedResult = content.reduce((agg, e) => [...e, ...agg], []);
-
-            console.info(`O365000002 Next collection in ${newState.poll_interval_sec} seconds`);
-            callback(null, flattenedResult, newState, newState.poll_interval_sec);
+        // Page aggregation handler
+        const handleContentCallback = ({parsedBody, nextPageUri}) => {
```
kkuzmin: I suggest to call it `listContentCallback`

collectors/o365/o365_collector.js
```diff
@@ -73,28 +78,63 @@ class O365Collector extends PawsCollector {
...
+            else{
+                return {
+                    parsedBody,
+                    nextPageUri
+                }
+            }
+        };
+
+        // If the state has a next page value, then just start with that.
+        const initialQuery = state.nextPage ?
```
kkuzmin: and `initialListContent`

collectors/o365/o365_collector.js
```diff
@@ -73,28 +78,63 @@ class O365Collector extends PawsCollector {
...
+
+            let newState;
+            if(nextPage === undefined){
+                newState = this._getNextCollectionState(state);
+            } else {
+                newState = this._getNextCollectionStateWithNextPage(state, nextPage);
+            }
+
+            console.info(`O365000002 Next collection for ${newState.stream} in ${newState.poll_interval_sec} seconds`);
+            callback(null, logs, newState, newState.poll_interval_sec);
```
kkuzmin: could you please add `return`. Same in catch clause.

collectors/o365/lib/o365_mgmnt/o365management.js
```diff
@@ -164,10 +165,11 @@ class O365Management extends msRestAzure.AzureServiceClient {
         }
         httpRequest.headers.set('Content-Type', 'application/json; charset=utf-8');
         httpRequest.body = null;
-        // Create handler
+
+        // Request Handler
         const handler = this.requestHandler(httpRequest);
         // Send Request
-        return client.sendRequest(httpRequest).then(handler);
+        return client.sendRequest(httpRequest).then(handler)
```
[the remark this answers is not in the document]
JDragovichAlertLogic: I have no idea. I'll put it back.

collectors/o365/lib/o365_mgmnt/o365management.js
```diff
@@ -81,11 +81,12 @@ class O365Management extends msRestAzure.AzureServiceClient {
                 error.response = msRest.stripResponse(res);
                 throw error
             }
-            return parsedBody;
+            const nextPageUri = headers.get('NextPageUri');
```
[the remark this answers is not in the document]
JDragovichAlertLogic: it returns `undefined`. I rely on that later for determining if we need to get the next page.

collectors/o365/o365_collector.js
```diff
@@ -30,9 +30,6 @@ const tsPaths = [
 
 console.log(process.env.O365_CONTENT_STREAMS)
```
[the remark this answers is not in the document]
JDragovichAlertLogic: yep

collectors/o365/o365_collector.js
```diff
@@ -49,13 +46,21 @@ class O365Collector extends PawsCollector {
...
-            since: startTs,
-            until: endTs,
-            poll_interval_sec: this.pollInterval
-        };
+
+        // Create a new 
+        const streams = JSON.parse(process.env.O365_CONTENT_STREAMS);
+        const initialStates = streams.map(stream => {
+            return {
+                stream,
```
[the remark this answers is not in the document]
JDragovichAlertLogic: yep, thats a shorthand for property assignment in ES6.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

collectors/o365/o365_collector.js
```diff
@@ -73,29 +77,63 @@ class O365Collector extends PawsCollector {
...
+                return {
+                    parsedBody,
+                    nextPageUri
+                }
+            }
+        };
+
+        // If the state has a next page value, then just start with that.
+        const initialListContent = state.nextPage ?
+            contentPromise =  m_o365mgmnt.getPreFormedUrl(state.nextPage):
```
kkuzmin: do we actually need this assignment?
I see `contentPromise` var is defined below.

collectors/o365/o365_collector.js
```diff
@@ -73,29 +77,63 @@ class O365Collector extends PawsCollector {
```
kkuzmin: Could you please also remove `collectorId: "none",` from `regValues`

kkuzmin: actually you can remove all of these values because PAWS will use them for registration for you.
But we probably want to add Azure tenant id and Azure audit streams into reg values.

collectors/o365/o365_collector.js
```diff
@@ -49,13 +46,21 @@ class O365Collector extends PawsCollector {
...
+
+        // Create a new 
+        const streams = JSON.parse(process.env.O365_CONTENT_STREAMS);
+        const initialStates = streams.map(stream => {
+            return {
+                stream,
+                since: startTs,
+                until: endTs,
+                nextPage: null,
+                poll_interval_sec: this.pollInterval
```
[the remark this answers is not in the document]
kkuzmin: please change it to 1 sec, because it is returned below

PR #41, opened 2020-01-10
Use constant paws_endpoint for o365

kkuzmin:
### Solution Description
Poll endpoint is constant for all o365 collectors. Therefore there is no need to have it as an input parameter in CFN.


kkuzmin:
The `paws_endpoint` env var is not actually used anywhere by o365 collector

PR #42, opened 2020-01-10
Fix initial states for Okta collector.

kkuzmin:
### Solution Description
handle all cases for initial states:
  1. **startDate** < now - pollInterval
  1. now - pollInterval < **startDate** < now
  1. **startDate** = now

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -31,13 +31,12 @@ class OktaCollector extends PawsCollector {
         const startTs = process.env.paws_collection_start_ts ? 
                 process.env.paws_collection_start_ts :
                     moment().toISOString();
-        const endTs = moment(startTs).add(this.pollInterval, 'seconds').toISOString();
-        const initialState = {
+        const initialState = this._getNextCollectionState({
             since: startTs,
-            until: endTs,
+            until: startTs,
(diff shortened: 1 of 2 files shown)


collectors/okta/okta_collector.js
```diff
@@ -31,13 +31,12 @@ class OktaCollector extends PawsCollector {
         const startTs = process.env.paws_collection_start_ts ? 
                 process.env.paws_collection_start_ts :
                     moment().toISOString();
-        const endTs = moment(startTs).add(this.pollInterval, 'seconds').toISOString();
-        const initialState = {
+        const initialState = this._getNextCollectionState({
             since: startTs,
-            until: endTs,
+            until: startTs,
```
OHaganA: from startTs to startTs? the initial state will have a 1 second time period?

OHaganA: Which of the 3 states does this refer to?

kkuzmin: This is a way to reuse `this._getNextCollectionState()` in order to get initial state. And `since: startTs` is not actually used by `_getNextCollectionState()` function.
The state 
```
{
            since: startTs,
            until: startTs
}
```
is sort of state 0 which is used as a state preceding the initial one

PR #46, opened 2020-01-15
okta token health check added

OHaganA:
### Problem Description
Okta api token can expire after 30 days, however the token is refreshed upon a successful API call. We needed a regular health check to ensure this token hasn't expired, as an expired token will not allow us to successfully call the okta endpoint to get logs.

### Solution Description
Implement a health check that involves a regular call to the okta endpoint. The call used is the getCurrentUser call, which is currently not implemented in the okta SDK for nodeJS, hence why I'm using the http call directly.


collectors/okta/health_checks.js
```diff
@@ -0,0 +1,41 @@
...
+
+ const https = require('https');
+ const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
+ 
+    function oktaTokenHealthCheck(asyncCallback){
+        /* No explicit call to validate an API token, the token expires after 30 days, this 30 day timeout is refreshed 
+         * upon any API calls. This call will refresh token, without being time consuming/destructive
+         */
+        var orgUrl = process.env.paws_endpoint;
+        var token = String.concat("SSWS", PawsCollector._pawsCreds.secret);
```
kkuzmin: I think a space char is missed after `SSWS`

collectors/okta/okta_collector.js
```diff
@@ -11,6 +11,7 @@
 
 const moment = require('moment');
 const okta = require('@okta/okta-sdk-nodejs');
+const healthChecks = require("./health_checks");
```
kkuzmin: please use single quotes

OHaganA:
This PR is subject to change as Okta are now looking to implement the getCurrentUser method in their SDK. https://github.com/okta/okta-sdk-nodejs/issues/126

PR #48, opened 2020-01-16
handle creds stored in s3 buckets

JDragovichAlertLogic:
### Problem Description
Google makes creds that are too big to go in lambda. se we need to put them in s3

### Solution Description
I tried to do this in a non-gross way. Not sure I succeeded. My head hurts.


cfn/paws-collector.template
```diff
@@ -85,6 +90,14 @@
             "AllowedPattern" : "(?:^\\d{4}(-\\d{2}){2}T(\\d{2}:){2}\\d{2}Z$)?"
         }
     },
+    "Conditions" : {
+        "CreateCredentialsBucket" : {"Fn::And": [
```
kkuzmin: S3 bucket may exist already

cfn/paws-collector.template
```diff
@@ -331,6 +390,8 @@
                                 "        Plaintext: event.ResourceProperties.Plaintext\n",
                                 "    };\n",
                                 "    const kms = new AWS.KMS();\n",
+                                "    const s3 = new AWS.S3();\n",
+                                "    // We will put the stuff in the bucket here tomorrow.\n",
```
kkuzmin: We will need some conditions. We don't want to put everything into the bucket, only `{"Fn::Equals" : [{"Ref" : "PawsAuthType"}, "s3object"]},`

cfn/paws-collector.template
```diff
@@ -388,6 +450,16 @@
                         "Arn"
                     ]
                 },
+                "AuthType": {
+                    "Ref": "PawsAuthType"
+                },
+                "CredsBucket": {
+                    "Fn:If": [
+                        "CreateCredentialsBucket",
+                        {"Fn::Sub" : ["${ImReallySorryIDidThisHack}", {"ImReallySorryIDidThisHack": {"Fn::GetAtt" : ["CredentialsBucket", "BucketName"]}}]},
```
kkuzmin: Why can't we have
```
"CredsBucket": {
                    "Ref": "PawsCredsBucket"
                },
```

cfn/paws-collector.template
kkuzmin: As we discussed we don't need to encrypt it when we put it into the bucket. S3 service will do this for you. 
We only need to encrypt the file which we going to store in /TMP

cfn/paws-collector.template
```diff
@@ -499,6 +571,9 @@
                       { "Ref" : "PawsCollectorTypeName" },
                       "-collector.json"
                   ]]},
+                  "aws_kms_arn": {
+                    "Fn::GetAtt": ["LambdaKmsKey", "Arn"]
```
kkuzmin: Not needed

paws_collector.js
```diff
@@ -17,15 +17,52 @@ const AWS = require('aws-sdk');
...
+                const encryptParams ={
+                    KeyId: process.env.aws_kms_arn,
+                    Plaintext: data.Body
+                };
+                kms.encrypt(encryptParams, (encryptError, encryptResponse) => {
+                    if (err) return reject(Error(err, err.stack));
+
+                    const fs.writeiFileSync(CREDS_FILE_PATH, encryptResponse.CiphertextBlob);
+                    // I know this is inefficient to pass the data I just encrypted to be immediately decrypted,
+                    // but I thought it was nicer to keep the subesquent function's inteface simple.
```
kkuzmin: This looks ok. You simply pass the encrypted content of a file as if you've read it.

paws_collector.js
```diff
@@ -47,17 +84,31 @@ class PawsCollector extends AlAwsCollector {
...
-                getDecryptedPawsCredentials(function(err, pawsCreds) {
-                    if (err){
-                        reject(err);
-                    } else {
-                        resolve({aimsCreds : aimsCreds, pawsCreds: pawsCreds});
-                    }
-                });
-            })
-        })
+                if(process.env.paws_auth_type === 's3object'){
```
kkuzmin: I think we can refactor this if statement to and call `getDecryptedPawsCredentials` only once.

cfn/paws-collector.template
```diff
@@ -270,6 +283,32 @@
                                 }
                             ]
                         },
+                        {
+                            "Effect":"Allow",
+                            "Action":"s3:CreateObject",
+                            "Resource":[
+                                {
+                                    "Fn::Join":[
```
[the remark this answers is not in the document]
JDragovichAlertLogic: yeps thats wrong. will fix

cfn/paws-collector.template
JDragovichAlertLogic: thats a really good point. i dont really know a good way of generating a good unique name.

cfn/paws-collector.template
JDragovichAlertLogic: yep absolutely. I'm going to check if it actually needs to put this stuff in there first.

cfn/paws-collector.template
JDragovichAlertLogic: My understandingis that the If function needs a string. this is a hack to turn that bucket name into a string. Though i think that is going to need some testing. the mysteries of cloud formation...

paws_collector.js
JDragovichAlertLogic: cool

cfn/paws-collector.template
```diff
@@ -79,6 +79,11 @@
             "Type": "String",
             "NoEcho": true
         },
+        "PawsCredsUri": {
+            "Description": "The S3 object where credentials are stored",
```
kkuzmin: Please add a note that it's only used together with `s3object` auth type

paws_collector.js
kkuzmin: Let's use `switch` statement instead of `if` here? To simplify adding new auth type in the future, like.
```
switch (process.env.paws_auth_type) {
  case 's3object':
    credsBuffer = getPawsCredsFile();
    break;
  default:
    credsBuffer = Buffer.from(process.env.paws_api_secret, 'base64');
}

getDecryptedPawsCredentials(credsBuffer);
```

cfn/paws-collector.template
```diff
@@ -270,6 +275,21 @@
...
                         },
+                        {
+                            "Effect":"Allow",
+                            "Action":"s3:Get*",
+                            "Resource":[
+                                {
+                                   "Fn::Join":["",
+                                      [
+                                         "arn:aws:s3:::",
+                                         { "Fn::Selecct": ["1", {"Fn:Split": ["//", {"Ref": "PawsCredsUri"}] } ] }
```
kkuzmin: typo `Selecct`

cfn/paws-collector.template
```diff
@@ -347,10 +352,11 @@
                                 "\n",
                                 "exports.handler = (event, context, callback) => {\n",
                                 "    if (event.ResourceType == 'AWS::CloudFormation::CustomResource' &&\n",
-                                "        event.RequestType == 'Create') {\n",
+                                "        event.RequestType == 'Create' &&\n",
+                                "        event.ResourceProperties.Plaintext.length  > 0) {\n",
                                 "        return encrypt(event, context);\n",
                                 "    }\n",
-                                "    return response.send(event, context, response.SUCCESS);\n",
+                                "    return response.send(event, context, response.SUCCESS, {EncryptedText: \"\"});\n",
```
kkuzmin: Why do we need this?

cfn/paws-collector.template
```diff
@@ -529,9 +535,15 @@
                   "paws_api_client_id":{
                       "Ref":"PawsClientId"
                   },
+                  "paws_kms_key_arn":{
+                      "Ref":"LambdaKmsKey"
```
kkuzmin: should we use `Fn::GetAttr` here?
```
"Fn::GetAtt": [
    "LambdaKmsKey",
    "Arn"
  ]
```

paws_collector.js
```diff
@@ -13,19 +13,59 @@
...
+                Bucket: s3PathParts.shift(),
+                Key: s3PathParts.join('/')
+            };
+            s3.getObject(params, (err, data) => {
+                if (err) return reject(new Error(err, err.stack));
+
+                // encrypt the file contents and cache on the lambda container file system
+                const encryptParams ={
+                    Plaintext: data.Body,
+                    KeyId: process.env.paws_kms_key_arn
```
kkuzmin: is it ok to pass ARN here? Or is it required to use key id instead?

paws_collector.js
```diff
@@ -47,17 +87,30 @@ class PawsCollector extends AlAwsCollector {
...
+
+                return credsPromise.then(credsBuffer => {
+                    getDecryptedPawsCredentials(credsBuffer, function(err, pawsCreds) {
+                        if (err){
+                            reject(err);
+                        } else {
+                            resolve({aimsCreds : aimsCreds, pawsCreds: pawsCreds});
+                        }
+                    });
+                }).catch(err => console.log("Error getting Paws Credentials", err));
```
kkuzmin: please use `console.err` and some error code.

test/paws_test.js
```diff
@@ -177,6 +202,67 @@ describe('Unit Tests', function() {
...
+        });
+
+        it('gets creds from s3 if present and writes them to cache', function(done){
+            const oldAuthType = process.env.paws_auth_type;
+            process.env.paws_auth_type = 's3object';
+            fileExistsStub = sinon.stub(fs, 'existsSync').callsFake(() => {
+                return false;
+            });
+
+            TestCollector.load().then(function({pawsCreds}){
```
kkuzmin: I'd be great to verify the value of `pawsCreds` in every test.

cfn/paws-collector.template
JDragovichAlertLogic: because that value is referenced later in the cfn template. It was throwinng an error if it wasn't being encrypted. i figured it was better to just return an empty string from this encrypt function than use a bunch of funky conditions in the cfn. You can encrypt an empy string with kms. It needs to have a minimum length of 1.

cfn/paws-collector.template
JDragovichAlertLogic: yep thats what we need. i misread my cfn cheatsheet, i thought that returned he arn

paws_collector.js
JDragovichAlertLogic: My tests have successfully encrypted the file using an arn.

test/paws_test.js
JDragovichAlertLogic: That is an issue that caused a lot of trouble when I was writing these tests...

The Paws collector class has some in-memory cacheing of the credentials, and my efforts to clear this cache after every test have been in vain. This cache means that the actual value of the creds cant be tested between tests. So we have to test whether the collecctor is at least going through the motions of using the file, even if its ultimatly going to use the value cached in memory at the end. 

This is apparently an issue with JS tests generally. I tries re-requireing the file and even tried using a library that claims to clear that cache. No dice.

So from what i can see, we would need to change the implementation fo that cache, or add method to clear it in order to test the creds value. Both of those options sounded like the tail wagging the dog, so I gave up trying.

JDragovichAlertLogic: This is the cache for reference.

https://github.com/alertlogic/paws-collector/blob/master/paws_collector.js#L20

cfn/paws-collector.template
```diff
@@ -594,6 +606,21 @@
...
                   },
+                  {
+                      "Effect":"Allow",
+                      "Action":"s3:Get*",
+                      "Resource":[
+                          {
+                             "Fn::Join":["",
+                                [
+                                   "arn:aws:s3:::",
+                                   { "Fn::Select": ["1", {"Fn::Split": ["//", {"Ref": "PawsCredsUri"}] } ] }
```
kkuzmin: Should be `PawsCredsS3Uri `

test/paws_test.js
kkuzmin: let's try to use `rewire` tool to clean the in memory cache value.

PR #49, opened 2020-01-16
Okta, O365, Auth: Fix make upload

kkuzmin:
### Solution Description
Upload all artifacts when running `AWS_LAMBDA_S3_BUCKET=some-test-bucket make upload`
Move collector templates under collector's folder.


collectors/template/cfn/collector.template.template
```diff
@@ -0,0 +1,102 @@
+{
```
JDragovichAlertLogic: If this file is being moved, the path in the template script needs to change as well.

https://github.com/alertlogic/paws-collector/blob/master/create-collector.js#L41

collectors/auth0/Makefile
```diff
@@ -26,7 +27,9 @@ deploy:
 	aws lambda update-function-code --function-name $(AWS_LAMBDA_FUNCTION_NAME) --zip-file fileb://$(AWS_LAMBDA_PACKAGE_NAME)
 
 upload:
-	aws s3 cp ./$(AWS_LAMBDA_PACKAGE_NAME) s3://$(AWS_LAMBDA_S3_BUCKET)
+	aws s3 cp ./$(AWS_LAMBDA_PACKAGE_NAME) s3://$(AWS_LAMBDA_S3_BUCKET)/packages/lambda/
```
JDragovichAlertLogic: Don't we need to push these packages to all the region specific buckets? Same with the other collector specific make files?

Makefile
```diff
@@ -25,6 +26,10 @@ publish:
 deploy:
 	aws lambda update-function-code --function-name $(AWS_LAMBDA_PAWS_FUNCTION_NAME) --zip-file fileb://$(AWS_LAMBDA_PAWS_PACKAGE_NAME)
 
+upload:
+	aws s3 cp ./$(AWS_LAMBDA_PACKAGE_NAME) s3://$(AWS_LAMBDA_S3_BUCKET)/packages/lambda/
```
JDragovichAlertLogic: Why would we need to upload a generic paws collector package?

collectors/auth0/Makefile
kkuzmin: Right. This is mostly done for test purposes when you'd like to upload artifact into your own private s3 bucket for testing.

Makefile
kkuzmin: right, we don't need to upload it.
But we still need to build it just incase someone runs it locally.

PR #54, opened 2020-01-30
readme cleanups

OHaganA:
### Problem Description
readme needed updating and linking to cwe-collector for key generation wasn't really helpful (in my opinion)
### Solution Description
Update readme


cfn/README-OKTA.md
```diff
@@ -8,18 +8,12 @@ Alert Logic OKTA Log Collector CloudFormation template.
 This procedure requires administrative permissions in AWS and your Alert Logic 
 Cloud Insight account. You also need to download the Alert Logic custom [CFT](https://github.com/alertlogic/paws-collector/blob/master/cfn/okta-collector.template).
```
OHaganA: Question about this line, we ask that they download the CFT but then in the deployment instructions below, we provide the s3 URL to the template. Which one is the actual process we recommend for customers installing their PAWS collector in their own environment? (assuming azcollect API will handle the install to the AL SaaS deployment)

PR #58, opened 2020-02-05
Okta: Report throttling errors

kkuzmin:
### Solution Description
Make okta collector report 429 rate exceeded errors.
Unfortunately okta SDK doesn't include any codes into errors, only test messages.

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -61,7 +61,7 @@ class OktaCollector extends PawsCollector {
             return callback(null, logAcc, newState, newState.poll_interval_sec);
         })
         .catch((error) => {
-            if (error.message && error.message.match(THROTTLING_ERROR_REGEXP)) {
+            if (this._isThrottlingError(error)) {
                 collector.reportApiThrottling(function() {
                     return callback(error);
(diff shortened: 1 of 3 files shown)


collectors/okta/okta_collector.js
```diff
@@ -60,7 +61,13 @@ class OktaCollector extends PawsCollector {
             return callback(null, logAcc, newState, newState.poll_interval_sec);
         })
         .catch((error) => {
-            return callback(error);
+            if (error.message && error.message.match(THROTTLING_ERROR_REGEXP)) {
```
JDragovichAlertLogic: Just for completeness, is there no status code for this sort of error? Do we need to do a match on this?

kkuzmin: Nope. SDK simply throws `new Error` in cae of throttling https://github.com/okta/okta-sdk-nodejs/blob/4643cf821c2aa77ecefbf92ddcf24e7c0b068862/src/default-request-executor.js#L93-L96

JDragovichAlertLogic: Fun. Ok sounds good to me.

PR #62, opened 2020-02-17
token healthcheck for okta

OHaganA:
### Problem Description
needed a health check to monitor status of okta collector

### Solution Description
add a healthcheck, calling the /me endpoint of the okta API manually as this method isn't implemented by the SDK

- [ ] No errors / warnings on build (including linter)
- [ ] 100% automated test coverage
- [ ] Source layout followed from other collectors
- [ ] No unnecessary code (debug logs, redundant comments, unused blocks of code, etc.)
- [ ] Logging of main steps in the collector’s code
- [ ] Logging of any requests that cause the collector to fail / throw an exception, with reason for failure / debugging info
- [ ] Registration/update/deregistration validated
- [ ] Stats and status validated
- [ ] CloudFormation template specific to setting up the collector, with documentation of any template parameters
- [ ] Collector README, including any set up caveats (both when installed in AL account and in customer’s account) – the user/team should be able to successfully set up, update, remove a collector following the README only
- [ ] Demo of set up, update, removal of a collector, with data shown to be correctly parsed in AL console search
- [ ] Links to API documentation
- [ ] At least temporary access to an environment where RCS can validate collector implementation
- [ ] Contact details in case access to this environment is needed in the future

### DIFF
--- a/collectors/okta/health_checks.js
+++ b/collectors/okta/health_checks.js
@@ -1,0 +1,41 @@
+/* -----------------------------------------------------------------------------
+ * @copyright (C) 2019, Alert Logic, Inc
+ * @doc
+ *
+ * Okta health checks.
+ *
+ * @end
+ * -----------------------------------------------------------------------------
+ */
+
+ 'use strict';
+
+ const https = require('https');
+ const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
+ 
+    function oktaTokenHealthCheck(asyncCallback){
+        /* No explicit call to validate an API token, the token expires after 30 days, this 30 day timeout is refreshed 
+         * upon any API calls. This call will refresh token, without being time consuming/destructive
+         */
+        var orgUrl = process.env.paws_endpoint;
+        var token = String.concat("SSWS ", PawsCollector._pawsCreds.secret);
+        var options = {
+            host: orgUrl,
+            path: "/api/v1/users/me",
+            headers: {
+                authorization: token
+            }
+        };
+        https.get(options, (res) => {
+          res.on('end', () => {
+                asyncCallback(null);
+          });
+      
+        }).on('error', (e) => {
+          asyncCallback(e);
+      });
+    }
+    
+module.exports = {
+    oktaTokenHealthCheck
+};
--- a/collectors/okta/health_checks.js
+++ b/collectors/okta/health_checks.js
@@ -11,19 +11,20 @@
  'use strict';
 
  const https = require('https');
- const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
  
     function oktaTokenHealthCheck(asyncCallback){
         /* No explicit call to validate an API token, the token expires after 30 days, this 30 day timeout is refreshed 
          * upon any API calls. This call will refresh token, without being time consuming/destructive
          */
         var orgUrl = process.env.paws_endpoint;
-        var token = "SSWS " + PawsCollector._pawsCreds.secret;
+        var token = "SSWS " + this._pawsCreds.secret;
         var options = {
-            host: orgUrl,
+            host: orgUrl.replace("https://", ""),
             path: "/api/v1/users/me",
             headers: {
-                authorization: token
+                Accept: "application/json",
+                Content: "application/json",
+                Authorization: token
             }
         };
         https.get(options, (res) => {
--- a/collectors/okta/health_checks.js
+++ b/collectors/okta/health_checks.js
@@ -18,7 +18,7 @@
          * upon any API calls. This call will refresh token, without being time consuming/destructive
          */
         var orgUrl = process.env.paws_endpoint;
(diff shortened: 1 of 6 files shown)


OHaganA:
@kkuzmin @JDragovichAlertLogic Just created a new PR, since rebasing the old one was getting painful, actioned the comments from the previous PR though

collectors/okta/health_checks.js
```diff
@@ -0,0 +1,41 @@
...
+                authorization: token
+            }
+        };
+        https.get(options, (res) => {
+          res.on('end', () => {
+                asyncCallback(null);
+          });
+      
+        }).on('error', (e) => {
+          asyncCallback(e);
```
JDragovichAlertLogic: I would add the Okta Collector Logging Prefix to this. and strinigify the result.

collectors/okta/health_checks.js
```diff
@@ -0,0 +1,41 @@
...
+
+ const https = require('https');
+ const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
+ 
+    function oktaTokenHealthCheck(asyncCallback){
+        /* No explicit call to validate an API token, the token expires after 30 days, this 30 day timeout is refreshed 
+         * upon any API calls. This call will refresh token, without being time consuming/destructive
+         */
+        var orgUrl = process.env.paws_endpoint;
+        var token = String.concat("SSWS ", PawsCollector._pawsCreds.secret);
```
JDragovichAlertLogic: this is fine, but as a poin you can just concat strings in JS with a `+` operator.

`"SSWS " + PawsCollector._pawsCreds.secret`

collectors/okta/okta_collector.js
```diff
@@ -28,6 +29,14 @@ const tsPaths = [
 
 class OktaCollector extends PawsCollector {
     
+    constructor(context, {aimsCreds, pawsCreds}){
+        super(context, 'paws',
+            AlAwsCollector.IngestTyps.LOGMSGS,
+            m_packageJson.version,
+            aimsCreds,
+            null, [healthChecks.oktaTokenHealthCheck], []);
+    }
```
JDragovichAlertLogic: This is not the constructor pattern for the PAWS collector. I think we'll need to modify the actual constructor of the paws collector to accept an array of health checks.

https://github.com/alertlogic/paws-collector/blob/master/paws_collector.js#L73-L83

collectors/okta/health_checks.js
```diff
@@ -0,0 +1,44 @@
...
+
+ const https = require('https');
+ 
+    function oktaTokenHealthCheck(asyncCallback){
+        /* No explicit call to validate an API token, the token expires after 30 days, this 30 day timeout is refreshed 
+         * upon any API calls. This call will refresh token, without being time consuming/destructive
+         */
+        var orgUrl = process.env.paws_endpoint;
+        console.log(orgUrl);
+        var token = "SSWS " + this._pawsCreds.secret;
```
kkuzmin: I suggest we avoid using `this` in a standalone function.

OHaganA:
Yep, got the PR to the lib here https://github.com/alertlogic/al-aws-collector-js/pull/43

collectors/okta/okta_collector.js
```diff
@@ -29,6 +30,12 @@ const tsPaths = [
 
 class OktaCollector extends PawsCollector {
 
+    constructor(context, {aimsCreds, pawsCreds}){
+        super(context,
+            {aimsCreds, pawsCreds},
+            [healthChecks.oktaTokenHealthCheck], []);
```
kkuzmin: I wonder if we can bind this function to `this` right here and avoid any changes in `al-aws-collector-js`. Also making `oktaTokenHealthCheck ` as a method of `OktaCollector` class

JDragovichAlertLogic: I don't really see the value of making it a class member. But that is more a personal style view than one with a particular technical objection.

It would be simpler to bind here. But I can see the benefit of making it a general solution in `al-aws-collector-js`. we will likely have more health checks that need to access the actual class for one resason or another, and i don't think we're going to have health checks that refer to the actual function instance themselves.

But I can see the value of not touching al-aws-collector-js this close to GA. So I'm cool with binding here.

paws_collector.js
```diff
@@ -71,12 +71,12 @@ class PawsCollector extends AlAwsCollector {
         });
     }
 
-    constructor(context, {aimsCreds, pawsCreds}) {
+    constructor(context, {aimsCreds, pawsCreds}, healthChecks, statsChecks) {
```
kkuzmin: please bump paws lib package version because constructor has changed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@alertlogic/paws-collector",
-  "version": "1.1.0",
```
kkuzmin: please change it to `1.1.3`
Also please update deps in okta collector

PR #64, opened 2020-02-17
Collector state batch splitting

daniel-greening:
### Problem Description
- SQS batch size is limited to 10 messages per batch

### Solution Description
- split batches into slices of >10 and submit concurrently to AWS

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -215,7 +215,6 @@ class PawsCollector extends AlAwsCollector {
     }
 
     _storeCollectionStateArray(pawsState, privCollectorStates, invocationTimeout, callback) {
-        // TODO: if 'privCollectorStates' length more than 10 split into multiple SQS messages batches (10 messages per batch)
         let collector = this;
         var sqs = new AWS.SQS({apiVersion: '2012-11-05'});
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -227,14 +226,31 @@ class PawsCollector extends AlAwsCollector {
...
             };
         });
-        
-        const params = {
-            Entries: SQSMsgs,
-            QueueUrl: process.env.paws_state_queue_url
-        };
+
+        var promises = [];
+        for (let i = 0; i < SQSMsgs.length; i += 10) {
```
kkuzmin: please define a constant for `10`

test/paws_test.js
```diff
@@ -131,7 +131,12 @@ class TestCollectorMultiState extends PawsCollector {
     }
     
     pawsGetLogs(state, callback) {
-        return callback(null, ['log1', 'log2'], [{state: 'new-state-1'}, {state: 'new-state-2'}], 900);
+        return callback(
+            null,
+            ['log1','log2'],
+            new Array(99).fill(0).map((e,i) => ({state: 'new-state-' + i})),
+            900
```
kkuzmin: It'd be nice to validate that `SQS.sendMessageBatch()` is called 10 times.

PR #70, opened 2020-03-02
fix local invocation stuff and makefile

JDragovichAlertLogic:
### Problem Description
Needed to update the local env stuff and the make sure the package was being built correctly.

### Solution Description
fix makefile and local env stuff

### DIFF
--- a/collectors/o365/package.json
+++ b/collectors/o365/package.json
@@ -9,9 +9,9 @@
     "test": "JUNIT_REPORT_PATH=./test/report.xml nyc --reporter=text mocha --colors"
   },
   "devDependencies": {
-    "aws-sdk": "*",
-    "aws-sdk-mock": "*",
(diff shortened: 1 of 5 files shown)


collectors/o365/package.json
```diff
@@ -9,9 +9,9 @@
     "test": "JUNIT_REPORT_PATH=./test/report.xml nyc --reporter=text mocha --colors"
   },
   "devDependencies": {
-    "aws-sdk": "*",
-    "aws-sdk-mock": "*",
-    "jshint": "^2.9.5",
+    "aws-sdk": "^2.624.0",
```
kkuzmin: would you like to bump package version for this change?

JDragovichAlertLogic: since those are dev dependencies, I don't think it's necessary. just trying very hard to banish the stars when I can.

PR #71, opened 2020-03-03
Add application Id to templates and log messages

daniel-greening:
### Problem Description
- application id required on log messages for better search parameters

### Solution Description
- add application id to log messages and make it configurable via templates


cfn/paws-collector.template
```diff
@@ -11,6 +11,10 @@
             "Type": "String",
             "NoEcho": true
         },
+        "AlApplicationId": {
+            "Description": "Alert Logic Application Id for collector logs",
+            "Type": "String"
+        },
```
kkuzmin: please default it to `paws`

collectors/auth0/auth0_collector.js
```diff
@@ -84,7 +84,8 @@ class Auth0Collector extends PawsCollector {
             priority: 11,
             progName: 'Auth0Collector',
             message: JSON.stringify(msg),
-            messageType: 'json/auth0'
+            messageType: 'json/auth0',
+            application_id: process.env.application_id
```
kkuzmin: I think we need to get away from collectors using `process.env` directly. Let's use base class properties, like
```
let collector = this;
...
application_id: collector.application_id
```

collectors/template/cfn/collector.template.template
```diff
@@ -11,6 +11,11 @@
             "Type": "String",
             "NoEcho": true
         },
+        "AlApplicationId": {
+            "Description": "Alert Logic Application Id for collector logs",
+            "Type": "String",
+            "Default": "collector_type"
```
kkuzmin: can we use `{{Type}}` here?

PR #73, opened 2020-03-05
fix application id

daniel-greening:
### Problem Description
- appending application ID fails due to collector state not being passed within formatLogs function

### Solution Description
- bind collector state to function


daniel-greening:
Closed in favor of https://github.com/alertlogic/paws-collector/pull/74

PR #77, opened 2020-03-09
remove apiToken from log object

daniel-greening:
### Problem Description
- Okta publish sensitive apiToken information in log objects 

### Solution Description
- Remove this field from the object for the time being

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -126,12 +126,10 @@ class OktaCollector extends PawsCollector {
 
         if (properties.length > 1) {
             if (!obj.hasOwnProperty(properties[0]))
-                return true
-
+                return
             return this.redactValue(properties.slice(1), obj[properties[0]])
         } else {
             obj[properties[0]] = undefined
-            return true
         }
     }
 }
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -26,6 +26,11 @@ const tsPaths = [
     { path: ['published'] }
 ];
 
+const sensitiveFields = [
+    "client.apiToken",
+    "client.http.defaultHeaders.Authorization"
+];
+
 
 class OktaCollector extends PawsCollector {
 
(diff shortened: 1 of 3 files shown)


collectors/okta/okta_collector.js
```diff
@@ -95,6 +95,9 @@ class OktaCollector extends PawsCollector {
         const ts = parse.getMsgTs(msg, tsPaths);
         const typeId = parse.getMsgTypeId(msg, typeIdPaths);
 
+        // TODO remove this when okta filter this field out
+        delete msg.client.apiToken;
```
kkuzmin: what happens if `msg.client` property doesn't exist?

OHaganA: Delete won't have any effect and just returns `true`

JDragovichAlertLogic: The api token is not just in that spot. It needs to be removed in the default http headers.

```
 "apiToken": "WHYUDOTHISOKTA?",
        "http": {
            "defaultHeaders": {
                "Authorization": "SSWS WHYUDOTHISOKTA?",
                "User-Agent": "@okta/okta-sdk-nodejs/3.1.0 node/11.10.0 darwin/18.7.0"
            },
```

collectors/okta/okta_collector.js
```diff
@@ -92,9 +97,14 @@ class OktaCollector extends PawsCollector {
     pawsFormatLog(msg) {
         let collector = this;
 
+        console.log("message = ", msg);
```
OHaganA: Stray console log here

collectors/okta/okta_collector.js
```diff
@@ -112,6 +122,21 @@ class OktaCollector extends PawsCollector {
...
+        let properties = Array.isArray(propertyPath) ? propertyPath : propertyPath.split(".")
+
+        if (properties.length > 1) {
+            if (!obj.hasOwnProperty(properties[0]))
+                return true
+
+            return this.redactValue(properties.slice(1), obj[properties[0]])
+        } else {
+            obj[properties[0]] = undefined
+            console.log("object", obj)
```
OHaganA: Same here with the console log

PR #78, opened 2020-03-09
Run all tests on every PR

JDragovichAlertLogic:
### Problem Description
We're only running base paws tests in the PR builder

### Solution Description
write a small script to run all tests and change travis config

- [ ] No errors / warnings on build (including linter)
- [ ] 100% automated test coverage
- [ ] Source layout followed from other collectors
- [ ] No unnecessary code (debug logs, redundant comments, unused blocks of code, etc.)
- [ ] Logging of main steps in the collector’s code
- [ ] Logging of any requests that cause the collector to fail / throw an exception, with reason for failure / debugging info
- [ ] Registration/update/deregistration validated
- [ ] Stats and status validated
- [ ] CloudFormation template specific to setting up the collector, with documentation of any template parameters
- [ ] Collector README, including any set up caveats (both when installed in AL account and in customer’s account) – the user/team should be able to successfully set up, update, remove a collector following the README only
- [ ] Demo of set up, update, removal of a collector, with data shown to be correctly parsed in AL console search
- [ ] Links to API documentation
- [ ] At least temporary access to an environment where RCS can validate collector implementation
- [ ] Contact details in case access to this environment is needed in the future


run-all-tests.sh
```diff
@@ -0,0 +1,6 @@
+dirs=$(find collectors/ -type d -maxdepth 1)
+echo $dirs
+for d in $dirs
+do
+    ( cd "$d" && echo "\n**********\n\nrunning tests for $d\n\n************\n\n" && npm run test )
```
OHaganA: This doesn't work for me unfortunately, it parses dirs as one big string. I just tried it manually in paws-collector dir
```
 dirs=$(find collectors/ -type d -maxdepth 1)
echo $dirs
for d in $dirs
do
    ( cd "$d" && echo "\n**********\n\nrunning tests for $d\n\n************\n\n" && npm run test )
done

collectors/
collectors//template
collectors//gsuite
collectors//o365
collectors//okta
collectors//auth0
collectors//googlestackdriver
cd: no such file or directory: collectors/\ncollectors//template\ncollectors//gsuite\ncollectors//o365\ncollectors//okta\ncollectors//auth0\ncollectors//googlestackdriver
```

JDragovichAlertLogic: hmmm thats odd.

are you in the project root?

run-all-tests.sh
```diff
@@ -0,0 +1,6 @@
+dirs=$(find collectors/ -type d -maxdepth 1)
```
kkuzmin: maybe we can have it in the Makefile?
And use `make -C collectors/okta`

JDragovichAlertLogic: That sounds like a good solution. Though I'm struggling to find a way to list all the directories in the collectors directory with make.

Makefile
```diff
@@ -16,6 +17,13 @@ compile: deps
 test: compile
 	npm run test
 	
+test-all: compile
+	npm run test
+	for d in $(COLLECTOR_DIRS); do \
+	    echo "\n************\n\nrunning tests for $$d\n\n************\n\n"; \
+	    cd $(shell pwd)/$$d && npm install && npm run test; \
```
kkuzmin: please change it to
```
make -C $$d test
```

Makefile
```diff
@@ -16,6 +17,13 @@ compile: deps
 test: compile
 	npm run test
 	
+test-all: compile
```
daniel-greening: Could we switch it so make test runs all tests and something like make test-framework runs base tests?

JDragovichAlertLogic: Wouldnt that run all tests when each collector runs `make test`? Since their make file is derived from the base make file?

kkuzmin: I'd prefer `make test` to run paws lib tests only. And `make test-collectors` to run all collectors tests

PR #81, opened 2020-03-13
Cfn template name fix

daniel-greening:
### Problem Description
- Names of gsuite, googlestackdriver and salesforce collector's cfn templates do not follow convention

### Solution Description
- Fix names to use `-` instead of `_`


daniel-greening:
Haven't been added to azcollect yet as far as I'm aware

PR #84, opened 2020-03-19
Salesforce v2

surajkadam:
Code refactored request to rest service Client.

### DIFF
--- a/collectors/salesforce/collector.js
+++ b/collectors/salesforce/collector.js
@@ -67,7 +67,7 @@ class SalesforceCollector extends PawsCollector {
         }
         const clientId = collector.clientId;
         const salesForceUser = process.env.paws_collector_param_string_1;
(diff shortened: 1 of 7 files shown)


collectors/salesforce/collector.js
```diff
@@ -14,7 +14,7 @@ const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
 const parse = require('@alertlogic/al-collector-js').Parse;
 const utils = require("./utils");
 var jwt = require('jsonwebtoken');
-var request = require('request');
+const RestServiceClient = require('@alertlogic/al-collector-js/al_util').RestServiceClient;
```
kkuzmin: In the latest version or `al-collector-js` this class is exported via `index.js`.
Could you please change it to `require('@alertlogic/al-collector-js').RestServiceClient`
Also please update deps in `package.json`

PR #85, opened 2020-03-19
Remove GSuite Paws Endpoint Parameter

daniel-greening:
### Problem Description
- No custom endpoint currently required for gsuite collectors 

### Solution Description
- remove PawsEndpoint parameter from cfn stack


collectors/gsuite/cfn/gsuite-collector.template
```diff
@@ -99,7 +95,7 @@
             "AlertlogicCustomerId" : { "Ref":"AlertlogicCustomerId" },
             "CollectorId" : { "Ref":"CollectorId" },
             "PollingInterval" : 60,
-            "PawsEndpoint" : { "Ref":"PawsEndpoint" },
+            "PawsEndpoint": "No-Endpoint",
```
kkuzmin: lets use gsuite specific endpoint which is used by the SDK

PR #87, opened 2020-03-19
Fix okta health checks

kkuzmin:
### Problem Description
Okta health check function was broken

### Solution Description
Fix okta health check function. 
Use `al-collector-js.RestServiceClient`.
Update tests.

### DIFF
--- a/collectors/okta/health_checks.js
+++ b/collectors/okta/health_checks.js
@@ -8,36 +8,37 @@
  * -----------------------------------------------------------------------------
  */
 
- 'use strict';
+'use strict';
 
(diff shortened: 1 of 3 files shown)


collectors/okta/health_checks.js
```diff
@@ -8,36 +8,37 @@
  * -----------------------------------------------------------------------------
  */
 
- 'use strict';
+'use strict';
 
- const https = require('https');
+const https = require('https');
```
JDragovichAlertLogic: We can remove this now since we're using the rest client.

PR #91, opened 2020-03-25
Update deps. Fix endpoint

kkuzmin:
### Problem Description
Sometimes customers input endpoint with `http` prefix

### Solution Description
Trim endpoint to domain name.
Update deps.

### DIFF
--- a/collectors/auth0/auth0_collector.js
+++ b/collectors/auth0/auth0_collector.js
@@ -39,7 +39,7 @@ class Auth0Collector extends PawsCollector {
 
     pawsGetLogs(state, callback) {
         let collector = this;
-        const hostname = process.env.paws_endpoint.toLowerCase().replace(HOSTNAME_REGEXP, '');
+        const hostname = process.env.paws_endpoint.replace(HOSTNAME_REGEXP, '');
         const auth0Client = new ManagementClient({
             domain: hostname,
             clientId: collector.clientId,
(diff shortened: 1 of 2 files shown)


collectors/auth0/auth0_collector.js
```diff
@@ -24,6 +24,8 @@ const tsPaths = [
     { path: ['date'] }
 ];
 
+const HOSTNAME_REGEXP = /^[https]*:\/\/|\/$/g;
```
mtrybulec: Why repeat the t here? A bit confusing...

kkuzmin: I left it for readability to highlight that we'd like to remove `http/https` prefixes.
Do you think it is confusing?

mtrybulec: :-) I guess it depends on the reader

collectors/auth0/auth0_collector.js
```diff
@@ -37,8 +39,9 @@ class Auth0Collector extends PawsCollector {
 
     pawsGetLogs(state, callback) {
         let collector = this;
+        const hostname = process.env.paws_endpoint.toLowerCase().replace(HOSTNAME_REGEXP, '');
```
mtrybulec: Hmm... might break if auth0 is case sensitive on endpoints; perhaps better to add i to the regex replace?

PR #92, opened 2020-03-25
Fix scope permissions

JDragovichAlertLogic:
### Problem Description
scopes are being fed by azcollect in JSON array format

### Solution Description
have collector handle this list of scopes.

### DIFF
--- a/collectors/gsuite/package.json
+++ b/collectors/gsuite/package.json
@@ -1,6 +1,6 @@
 {
     "name": "gsuite-collector",
-    "version": "1.1.6",
+    "version": "1.1.7",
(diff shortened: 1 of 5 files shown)


collectors/gsuite/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "gsuite-collector",
-    "version": "1.0.1",
+    "version": "1.0.2",
```
kkuzmin: gsuite is version 1.1.6 in our pipeline. could you please bring this up to speed

JDragovichAlertLogic: 👍

PR #95, opened 2020-03-27
Fix GSuite template json

daniel-greening:
### Problem Description
- malformed json in gsuite cfn template

### Solution Description
- fix json


daniel-greening:
Should probably add a linter to catch these issues

PR #102, opened 2020-04-03
add functionality to set param

JDragovichAlertLogic:
### Problem Description
Need to give the collector programmatically set its own api key param

### Solution Description
Implement that functionality.


paws_collector.js
```diff
@@ -117,7 +117,42 @@ class PawsCollector extends AlAwsCollector {
     prepareErrorStatus(errorString, streamName = 'none') {
         return super.prepareErrorStatus(errorString, streamName, this.pawsCollectorType);
     }
-    
+
+    setSecret(secretValue){
```
kkuzmin, line 121: let's name it `setPawsSecret` or `updatePawsSecret`

PR #104, opened 2020-04-07
Do not run collection if unregistered

kkuzmin:
### Problem Description
Paws collector continue collection even if registration fails.

### Solution Description
Make sure collection is not happening without `collector_id`.
Periodically reregister collector during checkin if it is unregistered.


kkuzmin:
test it

PR #106, opened 2020-04-10
O365 fix

JDragovichAlertLogic:
### Problem Description
Needed to start subscriptions on o365 streams in order to collect

### Solution Description
start streams during init and in every checkin, if they are not enabled.

### DIFF
--- a/collectors/o365/healthcheck.js
+++ b/collectors/o365/healthcheck.js
@@ -10,7 +10,6 @@
 
 'use strict';
 
-const health = require('@alertlogic/al-aws-collector-js').Health;
 const {
     listSubscriptions,
     startSubscription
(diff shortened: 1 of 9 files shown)


collectors/o365/healthcheck.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+function checkO365Subscriptions(callback){
+    return listSubscriptions()
+        .then(filterSubscriptions)
+        .then(filteredStreams => {
+            if(filteredStreams.length > 0){
+                console.log(`O365000101: Starting subscriptions for streams ${filteredStreams.join(', ')}`);
+            } else{
+                console.log(`O365000102: No streams need restarted.`);
```
kkuzmin, line 30: please use `console.info`

collectors/o365/o365_collector.js
```diff
@@ -33,7 +34,7 @@ const tsPaths = [
 class O365Collector extends PawsCollector {
 
     constructor(context, creds) {
-        super(context, creds, packageJson.version);
+        super(context, creds, packageJson.version, [checkO365Subscriptions], []);
     }
     
     pawsInitCollectionState(event, callback) {
```
kkuzmin, line 40: I think we need to check if subscription is started in `pawsInitCollectionState()`, and start it if disabled. Otherwise if subscription is disabled we'll have to wait for 15 mins until the next checkin trigger.

JDragovichAlertLogic: I do run the subscription at the end of this function. line 70-72.

I run the same healthcheck function. i just call it manually.

PR #113, opened 2020-04-21
Cisco AMP

surajkadam:
Alert Logic AWS Based API Poll (PAWS) Log Collector Library.

### Solution Description
This repository contains the AWS JavaScript Lambda function and CloudFormation Template (CFT) for deploying a log collector in AWS which will poll a 3rd party service API to collect and forward logs to the Alert Logic CloudInsight backend services.

### API Links and documentation.

Authentication - https://api-docs.amp.cisco.com/api_resources?api_host=api.amp.cisco.com&api_version=v1

>  Generating Client ID and API Key:

```
Log in to your AMP for Endpoints Console.
Go to Accounts > Business Settings.
Click Configure API Credentials under Features to generate the Client ID and secure API Key.
```
Audit logs - https://api-docs.amp.cisco.com/api_actions/details?api_action=GET+%2Fv1%2Faudit_logs&api_host=api.amp.cisco.com&api_resource=AuditLog&api_version=v1

- [x] No errors / warnings on build (including linter)
- [x] 92% automated test coverage
- [x] Source layout followed from other collectors
- [x] No unnecessary code (debug logs, redundant comments, unused blocks of code, etc.)
- [x] Logging of main steps in the collector’s code
- [x] Logging of any requests that cause the collector to fail / throw an exception, with reason for failure / debugging info
- [x] API throttling errors are understood and and properly handled
- [x] Registration/update/deregistration validated
- [x] CloudFormation template specific to setting up the collector, with documentation of any template parameters
- [x] Collector README, including any set up caveats (both when installed in AL account and in customer’s account) – the user/team should be able to successfully set up, update, remove a collector following the README only
- [x] Demo of set up, update, removal of a collector, with data shown to be correctly parsed in AL console search
- [x] Links to API documentation


collectors/ciscoamp/utils.js
```diff
@@ -0,0 +1,93 @@
...
+        function getData() {
+            if (pageCount < maxPagesPerInvocation) {
+                return restServiceClient.get(apiUrl, {
+                    headers: {
+                        "authorization": `Basic ${authorization}`
+                    },
+                    resolveWithFullResponse: true
+                }).then(({ headers, body }) => {
+                    pageCount++;
+                    if (parseInt(headers['x-ratelimit-remaining']) < 200) {
```
JDragovichAlertLogic, line 27: Why less than 200?

surajkadam: Hey Joe, Actually the hourly limit of request is 3000. After 2800 requests we check that condition and will stop requesting the server.

In AWS testing (Cloudwatch log groups), Every stack log streams AWS Assign an IP address and they are not the same. By IP our request limit is work (Every IP 3000 hourly request is allowed).

I think this condition rarely will work because As we know In hour 3600 seconds and we have 3000 request limit and every collector call at list required  2 to 3  seconds so in an hour max we used 1800 request with one log streams

PR #114, opened 2020-05-01
Add DD support in collectors

JDragovichAlertLogic:
- add DD support in paws framework
- add sample implementation in okta collector

### DIFF
--- a/cfn/paws-collector.template
+++ b/cfn/paws-collector.template
@@ -2,6 +2,16 @@
     "AWSTemplateFormatVersion": "2010-09-09",
     "Description": "Alert Logic template for creating a Poll based log collector",
     "Parameters": {
+        "DatadogApiKey" :{
+            "Description": "API Key for DataDog Metric collection. Use this OR DatadogKMSKey",
+            "Type": "String",
+            "Default": ""
+        },
+        "DatadogKMSKey" :{
+            "Description": "KMS ID for DataDog API Key. Use this OR DatadogApiKey",
+            "Type": "String",
+            "Default": ""
+        },
         "AlertlogicAccessKeyId": {
             "Description": "Alert Logic Access Key Id obtained from AIMS",
             "Type": "String"
@@ -98,16 +108,6 @@
             "Description": "The maximum number of pages of content data retrieved per invocation",
             "Type": "Number",
             "Default": 10
-        },
-        "DatadogApiKey" :{
-            "Description": "API Key for DataDog Metric collection. Use this OR DatadogKMSKey",
-            "Type": "String",
-            "Default": ""
-        },
-        "DatadogKMSKey" :{
-            "Description": "KMS ID for DataDog API Key. Use this OR DatadogApiKey",
-            "Type": "String",
-            "Default": ""
         },
         "CollectionStartTs": {
             "Description": "Optional. Timestamp when log collection starts. For example, 2020-01-13T16:00:00Z.",
--- a/cfn/paws-collector.template
+++ b/cfn/paws-collector.template
@@ -565,6 +565,12 @@
                       { "Ref" : "PawsCollectorTypeName" },
                       "-collector.json"
                   ]]},
+                  "DD_API_KEY": {
+                      "Ref": "DatadogApiKey"
+                  },
+                  "DD_KMS_API_KEY": {
+                      "Ref": "DatadogKMSKey"
+                  },
                   "al_application_id":{
                       "Ref":"AlApplicationId"
                   },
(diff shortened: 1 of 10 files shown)


test/paws_test.js
```diff
@@ -2,6 +2,7 @@ const assert = require('assert');
 const sinon = require('sinon');
 var AWS = require('aws-sdk-mock');
 const m_response = require('cfn-response');
+const ddLambda = require("datadog-lambda-js");
```
kkuzmin, line 5: please use single quotes

package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@alertlogic/al-aws-collector-js": "3.0.2",
```
kkuzmin, line 34: please bump package version

paws_collector.js
```diff
@@ -15,6 +15,7 @@ const debug = require('debug')('index');
 const AWS = require('aws-sdk');
 const fs = require('fs');
 const moment = require('moment');
+const ddLambda = require("datadog-lambda-js");
```
kkuzmin, line 18: please use single quotes

paws_collector.js
```diff
@@ -72,6 +73,15 @@ class PawsCollector extends AlAwsCollector {
         });
     }
 
+    // we can do whatever conditional handler wrapping here
+    static makeHandler(handlerFunc) {
+        if(process.env.DD_API_KEY || process.env.DD_KMS_API_KEY){
```
kkuzmin, line 78: we need to add these params to collector temaplates

collectors/okta/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "okta-collector",
-  "version": "1.1.5",
+  "version": "1.1.6",
```
kkuzmin, line 3: you also need to bump PAWS dependency version

paws_collector.js
```diff
@@ -122,6 +132,20 @@ class PawsCollector extends AlAwsCollector {
         return this._pawsHttpsEndpoint;
     };
 
+    reportDDMetric(name, value, tags = []) {
```
kkuzmin, line 135: should we have async version of it?

JDragovichAlertLogic: I'm not really sure if there is an async version the DD method exists.

We could add a callback to the paws method, but I'm not sure it would be useful.

collectors/okta/package.json
JDragovichAlertLogic: I guess we'll need to do this after we release the paws framework, otherwise the tests wont pass.

collectors/o365/package.json
```diff
@@ -24,6 +24,7 @@
     "@azure/ms-rest-js": "2.0.4",
     "@azure/ms-rest-azure-js": "2.0.1",
     "@azure/ms-rest-nodeauth": "2.0.5",
+    "datadog-lambda-js": "2.23.0",
```
kkuzmin, line 27: feels like this has to be a dependency for `paws-collector` lib only.

kkuzmin: Please also bump `paws-collector` dep version

collectors/okta/package.json
kkuzmin: so let's have a separate PR for `paws-collector` change.

paws_collector.js
kkuzmin: how are these env vars passed into a collector?

JDragovichAlertLogic: I thought that we could add these when azcollect creates a collector.

cfn/paws-collector.template
```diff
@@ -606,6 +616,12 @@
                   "paws_max_pages_per_invocation": {
                     "Ref":"PawsMaxPagesPerInvocation"
                   },
+                  "DD_API_KEY": {
```
kkuzmin, line 619: could you please move it on top, right after `aws_lambda_update_config_name`. I'd like to keep paws parameters close to each other.

paws_collector.js
```diff
@@ -122,6 +132,20 @@ class PawsCollector extends AlAwsCollector {
...
+        const baseTags = [
+            // some more tags here?
+            `paws_platform:${this.pawsCollectorType}`,
+            `collectorId:${this.application_id}`
```
kkuzmin, line 139: I guess it should be `applicationId`

cfn/paws-collector.template
```diff
@@ -99,6 +99,16 @@
             "Type": "Number",
             "Default": 10
         },
+        "DatadogApiKey" :{
```
kkuzmin, line 102: could you please move them to the top or bottom.

PR #115, opened 2020-05-04
Fix alps builds

kkuzmin:
### Solution Description
Fix ALPS tag publishing

### DIFF
--- a/ps_spec.yml
+++ b/ps_spec.yml
@@ -29,17 +29,18 @@ stages:
               
               PKGNAME=$(node -e 'console.log(require("./package.json").name)')
               PKGVERSION=$(node -e 'console.log(require("./package.json").version)')
-              V_VERSION=v$PKGVERSION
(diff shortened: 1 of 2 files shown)


ps_spec.yml
```diff
@@ -29,17 +29,18 @@ stages:
...
+              echo "https://$ALPS_GH_USER:$ALPS_GH_TOKEN@github.com" > ~/.git-credentials
+              git tag -a $PKGVERSION -m "$PKGNAME release $PKGVERSION" HEAD
+ 
               npm install
               
               if [ "$WORDCOUNT" -eq 0 ]; then
                  echo "PUBLISHING $PKGNAME $PKGVERSION"
                  npm run lint
-                 git push origin $V_VERSION
+                 git push origin $PKGVERSION
```
JDragovichAlertLogic, line 43: Dont oyu actually have to make the tag before you push?

`git tag -a -m 'Some message' '$PKGVERSION'`

kkuzmin: it is done at line 36

PR #121, opened 2020-05-15
Add stackname and check for handler

JDragovichAlertLogic:
Add stackname and check for handler

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -134,7 +134,6 @@ class PawsCollector extends AlAwsCollector {
         }
 
         const baseTags = [
-            // some more tags here?
             `paws_platform:${this.pawsCollectorType}`,
             `applicationId:${this.applicationId}`,
             `aws_account:${this._awsAccountId}`
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -128,10 +128,19 @@ class PawsCollector extends AlAwsCollector {
     };
 
     reportDDMetric(name, value, tags = []) {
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -128,10 +128,19 @@ class PawsCollector extends AlAwsCollector {
     };
 
     reportDDMetric(name, value, tags = []) {
+        // check if the API key is present. This will be a good proxy for if the handler is working
+        if (!process.env.DD_API_KEY){
+            return
+        }
+
+        // Distinguish between integration and prod collectors 
+        const stackName = process.env.al_api === "api.global-integration.product.dev.alertlogic.com" ? "integration" : "cd-us-production";
```
kkuzmin, line 137: Collectors are deployed in different AWS accounts. I don't think we need this.

JDragovichAlertLogic: So is there a way on the DD side to take the metrics coming form those accounts. right now the Integration metrics are showing up on the prod dashboards

paws_collector.js
```diff
@@ -128,10 +128,15 @@ class PawsCollector extends AlAwsCollector {
...
+        if (!process.env.DD_API_KEY){
+            return
+        }
+
         const baseTags = [
-            // some more tags here?
             `paws_platform:${this.pawsCollectorType}`,
-            `applicationId:${this.applicationId}`
+            `applicationId:${this.applicationId}`,
+            `aws_account:${this._awsAccountId}`
```
kkuzmin, line 139: let's use getter `this.aws_account_id`

PR #122, opened 2020-05-19
Bump dd metrics

JDragovichAlertLogic:
Bump paws collector package for dd metrics and add bump collector script.


package.json
```diff
@@ -17,6 +17,7 @@
     "lint": "jshint --exclude \"./node_modules/*\" **/*.js",
     "test": "JUNIT_REPORT_PATH=./test/report.xml nyc --reporter=cobertura mocha --colors",
     "create-collector": "node create-collector.js",
+    "bump-paws-version": "node update-paws-version.js",
```
kkuzmin, line 20: maybe, 'bump-collector-versions',
paws refers  to the library

Makefile
```diff
@@ -32,6 +32,9 @@ package.zip: node_modules/ *.js package.json
 publish:
 	npm run rel
 
+update-collector-versions:
+	npm run bump-collector-versions $(VERSION)
```
kkuzmin, line 36: so this `$(VERSION)` is a paws-collector library version, right?

kkuzmin: If so can we automatically get the latest version from npm or `paws-collector/package.json` and update it here.

JDragovichAlertLogic: We could. But i thought that would limit the utility of the script. I figured that we might want to roll a paws version back as well.

I could get the paws version from the latest package.json and use it in the script.

kkuzmin: > I figured that we might want to roll a paws version back as well

+1

> I could get the paws version from the latest package.json and use it in the script.

Let's have this as default behavior if `$VERSION` is not specified.

JDragovichAlertLogic: Cool. Beefed up the script. It takes the latest version from the main package.json if the argument is omitted. and it doesnt bump the collector version if it is already up to date.

PR #133, opened 2020-07-28
Sophos siem

surajkadam:
Alert Logic AWS Based API Poll (PAWS) Log Collector Library.

### Solution Description
This repository contains the AWS JavaScript Lambda function and CloudFormation Template (CFT) for deploying a log collector in AWS which will poll a 3rd party service API to collect and forward logs to the Alert Logic CloudInsight backend services.

- [x] No errors / warnings on build (including linter)
- [x] 93% automated test coverage
- [x] Source layout followed from other collectors
- [x] No unnecessary code (debug logs, redundant comments, unused blocks of code, etc.)
- [x] Logging of main steps in the collector’s code
- [x] Logging of any requests that cause the collector to fail / throw an exception, with reason for failure / debugging info
- [x] Registration/update/deregistration validated
- [x] CloudFormation template specific to setting up the collector, with documentation of any template parameters
- [x] Collector README, including any set up caveats (both when installed in AL account and in customer’s account) – the user/team should be able to successfully set up, update, remove a collector following the README only
- [x] Demo of set up, update, removal of a collector, with data shown to be correctly parsed in AL console search
- [x] Links to API documentation


collectors/sophossiem/test/sophossiem_test.js
```diff
@@ -0,0 +1,234 @@
...
+
+        it('Paws Init Collection State with start date behind 2 days Success', function (done) {
+            SophossiemCollector.load().then(function (creds) {
+                var collector = new SophossiemCollector(ctx, creds, 'sophossiem');
+                const startDate = moment().subtract(2, 'days').toISOString();
+                process.env.paws_collection_start_ts = startDate;
+                collector.pawsInitCollectionState({}, (err, initialStates, nextPoll) => {
+                    initialStates.forEach((state) => {
+                        assert.equal(state.poll_interval_sec, 1);
+                        assert.ok(state.from_date);
```
JDragovichAlertLogic, line 67: Could we test to make sure that the date is less than 24 hours in the past?

surajkadam: will do thanks

collectors/sophossiem/collector.js
```diff
@@ -0,0 +1,141 @@
...
+        else {
+            nextPollInterval = this.pollInterval;
+        }
+
+        return {
+            objectName: curState.objectName,
+            nextPage: nextPage,
+            poll_interval_sec: nextPollInterval
+        };
+    }
```
JDragovichAlertLogic, line 112: This new states doesn't seem to match the intial state thats set. Are we just going to rely on pagination?

What happens when we run out of pages? we'll need a new time based query right?

collectors/sophossiem/collector.js
```diff
@@ -0,0 +1,141 @@
...
+            "x-api-key": x_api_key,
+            "Authorization": clientSecret
+        };
+
+        const APIHostName = collector.pawsDomainEndpoint;
+        if (!APIHostName) {
+            return callback("The Host Name was not found!");
+        }
+
+        console.info(`SIEM000001 Collecting data for ${state.objectName}`);
```
JDragovichAlertLogic, line 85: I would prefer that we mark the time range of the query here as well. thats really helpful for determining if a collector is caught up or not.

collectors/sophossiem/collector.js
surajkadam: when request executes, it always gives us next_cursor value. 

The next cursor value is base64 encode string and when I decode this string it has timestamps.

If logs are more than limit then it will return has_more flag true.

collectors/sophossiem/collector.js
surajkadam: we need to decode that next_cursor value and extract timestamp value from that string.

JDragovichAlertLogic: It would be good if we could. How feasable do you think that is?

collectors/sophossiem/collector.js
JDragovichAlertLogic: Ok thats fair enough. as longa s we can always rely on the next crusor, we can use that.

collectors/sophossiem/collector.js
surajkadam: okay sure, I'll try

surajkadam: If logs available and pagination true then next_cursor return this string 
V2_CURSOR|2020-07-31T00:00:00.000Z|2020-07-31T08:47:38.849Z|es:[1596154817562,"asdsadasdasdasdasdasdas"]:10
V2_CURSOR|2020-07-31T00:00:00.000Z|2020-07-31T08:47:38.849Z|es:[1596156008047,"asdsadasssssasdACACASCS"]:10


if logs available and pagination false then next_cursor return this string 
V2_CURSOR|2020-07-31T08:55:51.970Z

if logs not available and pagination false then next_cursor return this string 
0|2020-07-31T08:57:11.071Z

surajkadam: Done

collectors/sophossiem/collector.js
JDragovichAlertLogic: Is it worth implementing some logic where if there is not next page token for some reason, that it just starts a time based state?

Maybe i'm being paranoid about it.

collectors/sophossiem/LICENSE
```diff
@@ -0,0 +1,8 @@
+Copyright 2019 ALERT LOGIC
```
kkuzmin, line 1: please fix the year

collectors/sophossiem/README.md
```diff
@@ -0,0 +1,105 @@
+# Sophossiem collector
+Alert Logic Sophossiem AWS Based API Poll (PAWS) Log Collector Library.
+
+# Overview
+This repository contains the AWS JavaScript Lambda function and CloudFormation 
+Template (CFT) for deploying a log collector in AWS which will poll 3rd party service API to collect and 
+forward logs to the Alert Logic CloudInsight backend services.
```
kkuzmin, line 7: please mention which Sophos API is used for getting data from

PR #138, opened 2020-08-06
Cb split templates

JDragovichAlertLogic:
Adding themis template validators for the two carbonblack collectors. these will replace the existing themis validator.


JDragovichAlertLogic:
test it

JDragovichAlertLogic:
test it

PR #139, opened 2020-08-06
update package.json

JDragovichAlertLogic:
### Problem Description
Dependabot suggested that we need to upgrade the auth0 library

### Solution Description
Update dep


kkuzmin:
test it

PR #141, opened 2020-08-10
O365 moment state issue

surajkadam:
O365 moment state issue

### DIFF
--- a/collectors/o365/o365_collector.js
+++ b/collectors/o365/o365_collector.js
@@ -85,11 +85,11 @@ class O365Collector extends PawsCollector {
         let collector = this;
 
         if (!moment(state.since).isValid() || !moment(state.until).isValid() || state.since === undefined || state.until === undefined) {
-            const newStartTs = moment();
-            state.since = newStartTs.toISOString();
-            state.until = newStartTs.add(collector.pollInterval, 'seconds').toISOString();
+            const { nextUntilMoment, nextSinceMoment, nextPollInterval } = calcNextCollectionInterval('hour-day-progression', moment(), this.pollInterval);
+            state.since = nextSinceMoment.toISOString();
+            state.until = nextUntilMoment.toISOString();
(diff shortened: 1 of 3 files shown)


collectors/o365/o365_collector.js
```diff
@@ -83,6 +83,16 @@ class O365Collector extends PawsCollector {
 
     pawsGetLogs(state, callback) {
         let collector = this;
+
+        if (!moment(state.since).isValid() || !moment(state.until).isValid() || state.since === undefined || state.until === undefined) {
+            const newStartTs = moment();
+            state.since = newStartTs.toISOString();
+            state.until = newStartTs.add(collector.pollInterval, 'seconds').toISOString();
+            state.nextPage = null;
+            state.poll_interval_sec = collector.pollInterval;
```
JDragovichAlertLogic, line 92: I would just use the built in progressions that we use elsewhere in the collector.

https://github.com/alertlogic/paws-collector/blob/master/collectors/o365/o365_collector.js#L165

collectors/o365/test/o365_test.js
```diff
@@ -401,6 +401,27 @@ describe('O365 Collector Tests', function() {
...
         });
 
+        it('check if state moment is null or blank', function (done) {
+            setO365MangementStub();
+            O365Collector.load().then(function (creds) {
+                var collector = new O365Collector(ctx, creds, 'o365');
+                const curState = {
+                    stream: "FakeStream",
+                    since: "",
+                    until: null,
```
JDragovichAlertLogic, line 411: I would add a test case where both since and until are undefined

collectors/o365/o365_collector.js
surajkadam: Done

collectors/o365/test/o365_test.js
surajkadam: done

PR #144, opened 2020-08-31
remove check for domain url.

JDragovichAlertLogic:
### Problem Description
There was a check that looked fo rhte domain url, this is dead code because this url is not needed. We need to remove the check to support collectors being created without the domain url

### Solution Description
remove url check


kkuzmin (approved):

surajkadam (changes requested):
Please remove the Sophos endpoint from the CFN template. [](https://github.com/JDragovichAlertLogic/paws-collector/blob/sophos-fix/collectors/sophos/cfn/sophos-collector.template#L54-L58)

PR #148, opened 2020-09-29
Fix o365

kkuzmin:
### Problem Description
Status field isn't set properly is O365 subscription check reports an error.

### Solution Description
Stringify the error and returned in the health check callback.


JDragovichAlertLogic (approved):
LGTM. thanks Kanstatnsin

OHaganA (approved):

PR #149, opened 2020-10-13
ENG-14517 - [salesforce] remediation for failing to query inaccessible objects

mohasin-nadaf:
Some salesforce customers may not have the platform security events enabled,
In such cases have a relevant error status for these problems with details on which objects are failing to be queried.


mohasin-nadaf:
@daniel-greening Can you let me know this looks good or still missing something.

PR #150, opened 2020-10-13
Sophos handled errors for organization type and request limit.

surajkadam:
Sophos handled errors for organization type and request limit.


JDragovichAlertLogic (changes requested):
Looks good Suraj. But could we get some units tests for this?

JDragovichAlertLogic (approved):
Looks good to me. thanks Suraj

PR #153, opened 2020-10-21
update al-aws-collector dependency with unlink file fix

daniel-greening:
### Problem Description
- Collection periodically fails due to unlink of cache throwing a ENOENT error: https://github.com/alertlogic/al-collector-js/pull/42

### Solution Description
- Update al-aws-collector dependency to include fix


JDragovichAlertLogic (approved):
Looks good to me. thanks dan

PR #157, opened 2020-10-28
ENG-16247 : Additional collector metrics for client errors

rakhimundhada15:
- Added the method to report client errors on ddMetrics and cloud watch.
-Added the method reportClientOk to  report client execute successfully ,So we can check how often 3rd party APIs get called

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -444,8 +444,9 @@ class PawsCollector extends AlAwsCollector {
     /**
      * Report the client errors and show case on DDMetrics and cloudwatch
      * @param callback 
+     * @param errorCode - error type
      */
-    reportClientErrors(callback) {
+    reportClientErrors(errorCode = 'unknown', callback) {
         var cloudwatch = new AWS.CloudWatch({apiVersion: '2010-08-01'});
         const params = {
             MetricData: [
@@ -468,7 +469,7 @@ class PawsCollector extends AlAwsCollector {
             ],
             Namespace: 'PawsCollectors'
         };
-        this.reportDDMetric("client_errors", 1);
+        this.reportDDMetric(`${errorCode}_client_errors`, 1);
         return cloudwatch.putMetricData(params, callback);
     };
     
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -378,7 +378,17 @@ class PawsCollector extends AlAwsCollector {
                 return collector.checkStateSqsMessage(stateSqsMsg, asyncCallback);
             },
             function(asyncCallback) {
-                return collector.pawsGetLogs(pawsState.priv_collector_state, asyncCallback);
+                return collector.pawsGetLogs(pawsState.priv_collector_state, (err, ...remainingParams) => {
+                    if (err) {
+                        collector.reportClientError(err, () => {
+                            return asyncCallback(err);
+                        });
+                    } else {
+                        collector.reportClientOK(() => {
+                            return asyncCallback(null, ...remainingParams)
+                        });
+                    }
+                });
             },
             function(logs, privCollectorState, nextInvocationTimeout, asyncCallback) {
                 console.info('PAWS000200 Log events received ', logs.length);
@@ -469,12 +479,8 @@ class PawsCollector extends AlAwsCollector {
             ],
             Namespace: 'PawsCollectors'
         };
-        // Tags can be up to 200 characters long
-        let errorMesage = error.errorMessage? `error_message:${error.errorMessage}`:`error_message:not_define`;
-        if(errorMesage.length > 200){
-             errorMesage = errorMesage.slice(0,199);
-        }
-        this.reportDDMetric("client", 1, [`result:error`,`error_code:${error.errorCode}`,errorMesage]);
+        let errorCode = typeof (error) === 'object' && error.errorCode ? error.errorCode : 'unknown';
+        this.reportDDMetric("client", 1, [`result:error`,`error_code:${errorCode}`]);
         return cloudwatch.putMetricData(params, callback);
     };
 
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -444,14 +444,51 @@ class PawsCollector extends AlAwsCollector {
     /**
      * Report the client errors and show case on DDMetrics and cloudwatch
      * @param callback 
-     * @param errorCode - error type
+     * @param error 
      */
-    reportClientErrors(errorCode = 'unknown', callback) {
+    reportClientError(error, callback) {
+        var cloudwatch = new AWS.CloudWatch({ apiVersion: '2010-08-01' });
+        const params = {
+            MetricData: [
+                {
+                    MetricName: "PawsClientError",
+                    Dimensions: [
+                        {
+                            Name: 'CollectorType',
+                            Value: this._pawsCollectorType
+                        },
+                        {
+                            Name: 'FunctionName',
+                            Value: process.env.AWS_LAMBDA_FUNCTION_NAME
+                        }
+                    ],
+                    Timestamp: new Date(),
+                    Unit: 'Count',
+                    Value: 1
+                }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        // Tags can be up to 200 characters long
+        let errorMesage = error.errorMessage? `error_message:${error.errorMessage}`:`error_message:not_define`;
+        if(errorMesage.length > 200){
+             errorMesage = errorMesage.slice(0,199);
+        }
+        this.reportDDMetric("client", 1, [`result:error`,`error_code:${error.errorCode}`,errorMesage]);
+        return cloudwatch.putMetricData(params, callback);
+    };
+
+    /**
+    * Collector to report client execute successfully, 
+    * So we can check how often 3rd party APIs get called.
+    * @param callback 
+    */
+    reportClientOK( callback) {
         var cloudwatch = new AWS.CloudWatch({apiVersion: '2010-08-01'});
         const params = {
             MetricData: [
               {
-                MetricName: "PawsClientErrors",
+                MetricName: "PawsClientOK",
                 Dimensions: [
                   {
                     Name: 'CollectorType',
@@ -469,7 +506,7 @@ class PawsCollector extends AlAwsCollector {
             ],
             Namespace: 'PawsCollectors'
         };
-        this.reportDDMetric(`${errorCode}_client_errors`, 1);
+        this.reportDDMetric("client", 1, [`result:ok`]);
         return cloudwatch.putMetricData(params, callback);
     };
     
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -440,6 +440,38 @@ class PawsCollector extends AlAwsCollector {
...
         return cloudwatch.putMetricData(params, callback);
     };
+
+    /**
+     * Report the client errors and show case on DDMetrics and cloudwatch
+     * @param callback 
+     * @param errorCode - error type
+     */
+    reportClientErrors(errorCode = 'unknown', callback) {
+        var cloudwatch = new AWS.CloudWatch({apiVersion: '2010-08-01'});
```
daniel-greening, line 450: Do the collectors have permission to report Cloudwatch metrics?

paws_collector.js
```diff
@@ -440,6 +440,38 @@ class PawsCollector extends AlAwsCollector {
...
+                ],
+                Timestamp: new Date(),
+                Unit: 'Count',
+                Value: 1
+              }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        this.reportDDMetric(`${errorCode}_client_errors`, 1);
+        return cloudwatch.putMetricData(params, callback);
```
daniel-greening, line 473: I'm not sure the interface for this function should be coupled with what AWS cloud watch returns

paws_collector.js
```diff
@@ -440,6 +440,38 @@ class PawsCollector extends AlAwsCollector {
...
+                  }
+                ],
+                Timestamp: new Date(),
+                Unit: 'Count',
+                Value: 1
+              }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        this.reportDDMetric(`${errorCode}_client_errors`, 1);
```
daniel-greening, line 472: Could you please rename to `client` and have the following tags: 

- result: error
- reason: ERRORCODE 

So call should look something like:
``` javascript
this.reportDDMetric("client", 1, [`result:error`,`error_code:${errorCode}`])
```

paws_collector.js
```diff
@@ -440,6 +440,38 @@ class PawsCollector extends AlAwsCollector {
         this.reportDDMetric('api_throttling', 1)
         return cloudwatch.putMetricData(params, callback);
     };
+
+    /**
+     * Report the client errors and show case on DDMetrics and cloudwatch
+     * @param callback 
+     * @param errorCode - error type
+     */
+    reportClientErrors(errorCode = 'unknown', callback) {
```
daniel-greening, line 449: given that we're reporting a single error metric, perhaps `reportClientError` is a better name here.

Also, I think just passing a string here limits us in what clients are able to report. Using the `Error` object in Node might be a better interface for this as we can pull code from there: https://nodejs.org/api/errors.html#errors_error_code

@kkuzmin  thoughts?

daniel-greening (changes requested):
Couple of comments on this one. 

Also could you implement a means for collectors to report client executed successfully so we can better gauge how often we're calling 3rd party APIs. Something like:

```javascript
reportClientOk(callback)
```

where you report the metric using datadog like so: 

```javascript
this.reportDDMetric("client", 1, [`result:ok`])
```

paws_collector.js
kkuzmin: +1 for the object interfaces

JDragovichAlertLogic (changes requested):

paws_collector.js
rakhimundhada15: updated the reportDDMetric parameters as this.reportDDMetric("client", 1, [`result:error`,`error_code:${error.errorCode}`,errorMesage]);

paws_collector.js
rakhimundhada15: updated to error object from string

paws_collector.js
```diff
@@ -440,6 +440,75 @@ class PawsCollector extends AlAwsCollector {
         this.reportDDMetric('api_throttling', 1)
         return cloudwatch.putMetricData(params, callback);
     };
+
+    /**
+     * Report the client errors and show case on DDMetrics and cloudwatch
+     * @param callback 
+     * @param error 
+     */
+    reportClientError(error, callback) {
```
daniel-greening, line 449: Just so I'm clear on work left to be done here. Is the intention for subclasses of paws-collector to call these methods? 

If so I don't think that's the right approach. I wouldn't want this behaviour to be done at the subclass level, IMO we'd want to call this function [here](https://github.com/alertlogic/paws-collector/blob/bc67a70c8463e9c05037a54982a000c3d9dd69a2/paws_collector.js#L380-L382):

```javascript
function(asyncCallback) {
    return collector.pawsGetLogs(pawsState.priv_collector_state, asyncCallback);
}
```

@JDragovichAlertLogic  @kkuzmin thoughts?

JDragovichAlertLogic: I think that because each collector si going to have a different schema for errors (ie. they might be formmated by an SDK) we will need each collector to do its own logic ion formatting errors.

I don't know if there is going to be a library level solution fo this issue. For example, in the o365 collector, we're pulling from two different apis in sequence through an SDK.

We could call this method in the same way we call the throttling metric method.

https://github.com/alertlogic/paws-collector/blob/bc67a70c8463e9c05037a54982a000c3d9dd69a2/collectors/okta/okta_collector.js#L79-L85

I mean we could handle this on the base class level. but we would need to force each collector to implement a `handleClientError` method because i dont really thing we can trust thee sdks to give us a raw nodejs error.

So a reasonable example is that we have funky error codes coming back from the okta api that we might want to capture that give us more info than just the http error. Not sure how the sdk handles this.

https://developer.okta.com/docs/reference/error-codes/

paws_collector.js
```diff
@@ -440,6 +440,75 @@ class PawsCollector extends AlAwsCollector {
...
+                    Value: 1
+                }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        // Tags can be up to 200 characters long
+        let errorMesage = error.errorMessage? `error_message:${error.errorMessage}`:`error_message:not_define`;
+        if(errorMesage.length > 200){
+             errorMesage = errorMesage.slice(0,199);
+        }
```
JDragovichAlertLogic, line 476: Same here. we don't want to be calling length and slice on something that is not a string. It'll throw an error.

paws_collector.js
JDragovichAlertLogic: Ok after some kibitzing offline, Dan and I hashed out  plan for how this can be implemented.

Dan is correct that we should define a reasonable shape for this error object, and it should probably resemble the standard nodejs error.

However we should call this reporting method after the logs are collected and rely on the `asyncCallback` in `pawsGetLogs` to pass the error in. Each collector will need to make sure hte error object is the correct shape.

somthing like this:

``` javascript
function(asyncCallback) {
        return collector.pawsGetLogs(pawsState.priv_collector_state, (err, ...remainingParams) => {
      if (err) {
        this.reportClientError(err);
        asyncCallback(err);
      } else {
        this.reportClientOk();
        asyncCallback(null, ...remainingParams)
      }
    });
},
```

JDragovichAlertLogic (changes requested):

paws_collector.js
```diff
@@ -440,6 +440,75 @@ class PawsCollector extends AlAwsCollector {
...
+                    ],
+                    Timestamp: new Date(),
+                    Unit: 'Count',
+                    Value: 1
+                }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        // Tags can be up to 200 characters long
+        let errorMesage = error.errorMessage? `error_message:${error.errorMessage}`:`error_message:not_define`;
```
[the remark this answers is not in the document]
daniel-greening: I'd do away with sending these errorMessages to datadog. We pay based on unique tags so could end up getting a massive bill here if we're not careful.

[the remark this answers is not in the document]
rakhimundhada15: I have removed error message and just sending error code to DDMetric as  `this.reportDDMetric("client", 1, [`result:error`,`error_code:${errorCode}`]);`

paws_collector.js
rakhimundhada15: Previously I am sending message in string format from each collector , but now I have remove message parameter from tags.

paws_collector.js
```diff
@@ -440,6 +450,71 @@ class PawsCollector extends AlAwsCollector {
...
+                        }
+                    ],
+                    Timestamp: new Date(),
+                    Unit: 'Count',
+                    Value: 1
+                }
+            ],
+            Namespace: 'PawsCollectors'
+        };
+        let errorCode =  (typeof(error)=== 'object' &&  error.errorCode) ? error.errorCode:'unknown';
```
JDragovichAlertLogic, line 482: formatting issue. space around `===`

JDragovichAlertLogic: also space around `:` in ternary statement

paws_collector.js
```diff
@@ -440,6 +450,71 @@ class PawsCollector extends AlAwsCollector {
...
+        this.reportDDMetric("client", 1, [`result:error`,`error_code:${errorCode}`]);
```
JDragovichAlertLogic, line 483: I would still call this "client_error" and distinguish it from successful client calls, rather than just relying on tags.

@daniel-greening what do you think?

paws_collector.js
JDragovichAlertLogic: also, i don't think that you need the parens around teh conditional statement, but it is fine i think.

JDragovichAlertLogic (changes requested):
Thanks Rakhi. A few style issues then we're ok to merge

paws_collector.js
rakhimundhada15: I will formate and update

paws_collector.js
daniel-greening: No real preference, though what's wrong with relying on tags?

JDragovichAlertLogic: nothing I guess. I thought i would have been easier to build graphs off of. But that might just be my own way of doing things.

JDragovichAlertLogic (approved):
Looks good. Thanks Rakhi

paws_collector.js
```diff
@@ -378,7 +378,17 @@ class PawsCollector extends AlAwsCollector {
                 return collector.checkStateSqsMessage(stateSqsMsg, asyncCallback);
             },
             function(asyncCallback) {
-                return collector.pawsGetLogs(pawsState.priv_collector_state, asyncCallback);
+                return collector.pawsGetLogs(pawsState.priv_collector_state, (err, ...remainingParams) => {
```
daniel-greening, line 381: Should probably add tests for this new flow

daniel-greening (changes requested):

paws_collector.js
rakhimundhada15: For now I have added the test cases for newly added methods .I will check if i can add more test cases to test flow after my vacation.

PR #158, opened 2020-10-28
Limit concurrent calls to o365 Management API 

daniel-greening:
### Problem Description
- When collecting a large number of objects, we hit the OS socket limit due to large number of parallel calls to o365 api's

### Solution Description
- limit concurrent api calls to 20

### DIFF
--- a/collectors/o365/o365_collector.js
+++ b/collectors/o365/o365_collector.js
@@ -17,6 +17,7 @@ const { checkO365Subscriptions } = require('./healthcheck');
 const packageJson = require('./package.json');
 
 const parse = require('@alertlogic/al-collector-js').Parse;
+const asyncPool = require("tiny-async-pool");
 
 // Subtracting less than 7 days to avoid weird race conditions with the azure api...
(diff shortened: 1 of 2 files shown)


mohasin-nadaf (approved):

JDragovichAlertLogic (changes requested):
Thanks for this Dan. This will work in the short term.

we probably want to limit the number of paralel executions for getting content as well. We can use paralle limit wit teh async library, it might be better than rolling out implentation with promises.

https://caolan.github.io/async/v3/docs.html#parallelLimit

JDragovichAlertLogic (approved):
Looks good. but it might be useful to have a test for this as well.

daniel-greening:
What would be testing?

JDragovichAlertLogic:
I'm thinking it might be good that we put some sort of check in there to make sure that we limit the concurrency in this collector because of the way the MS api is. But I cant really think of a good way of doing that without yoking it to a particular implmentation.

So yeah +1. lets just ship it.

PR #159, opened 2020-11-11
ENG-17586 [All-paws-collectors] Stream specific status reporting

mohasin-nadaf:
Prerequisite:
https://github.com/alertlogic/paws-collector/pull/172


paws_collector.js
```diff
@@ -96,14 +96,14 @@ class PawsCollector extends AlAwsCollector {
         }
     }
 
-    constructor(context, {aimsCreds, pawsCreds}, childVersion, healthChecks = [], statsChecks = []) {
+    constructor(context, {aimsCreds, pawsCreds}, childVersion,streams =[], healthChecks = [], statsChecks = []) {
```
kkuzmin, line 99: please exctract `paws_collector` lib changes into separate PR. Later on you'll need to update `paws-collector` dependency in collectors' package.json's

collectors/auth0/auth0_collector.js
```diff
@@ -27,15 +27,17 @@ const tsPaths = [
 
 const HOSTNAME_REGEXP = /^[htps]*:\/\/|\/$/gi;
 
+const STREAM_NAME = 'default';
 
 class Auth0Collector extends PawsCollector {
 
     constructor(context, creds) {
-        super(context, creds, packageJson.version);
+        super(context, creds, packageJson.version,[]);
```
kkuzmin, line 35: `[]` is not needed because it is already defaulted in the constructor

collectors/auth0/auth0_collector.js
```diff
@@ -27,15 +27,17 @@ const tsPaths = [
...
     }
     
     pawsInitCollectionState(event, callback) {
         const initialState = {
+            stream : STREAM_NAME,
```
kkuzmin, line 40: not sure why we need default stream here as we already passed `[]` to the constructor.

collectors/carbonblack/collector.js
```diff
@@ -25,16 +25,16 @@ let tsPaths = [];
 class CarbonblackCollector extends PawsCollector {
 
     constructor(context, creds) {
-        super(context, creds, packageJson.version);
+        super(context, creds, packageJson.version,process.env.paws_collector_param_string_1);
```
kkuzmin, line 28: I suggest we validate `paws_collector_param_string_1` before passing it to the constructor.

collectors/auth0/auth0_collector.js
```diff
@@ -27,15 +27,17 @@ const tsPaths = [
 
 const HOSTNAME_REGEXP = /^[htps]*:\/\/|\/$/gi;
 
+const STREAM_NAME = 'default';
```
kkuzmin, line 30: we need to extract this constant into the library because it is used in multiple collectors.

kkuzmin (changes requested):

collectors/auth0/auth0_collector.js
rakhimundhada15: sure I will update this.

paws_collector.js
rakhimundhada15: Okay I will raise separate pr with latest version of al-aws-collector and bump the paws version.

collectors/auth0/auth0_collector.js
rakhimundhada15: I will also think now it will not needed but when we had discussion ,if their is no stream set to default so that why I have added. I will remove this.

collectors/auth0/auth0_collector.js
rakhimundhada15: I have removed default value for stream and it will post the status with collector type  (ex paws_okta_status) as it handle d in Paws and al-aws-collector while [post error status](https://github.com/alertlogic/al-aws-collector-js/pull/62/files#diff-c06f5f0e08865e628912333b6bb7941cf34fcc02b6fc4163a80e350b5bfd4c47R188) and [post ok status](https://github.com/alertlogic/al-aws-collector-js/pull/62/files#diff-c06f5f0e08865e628912333b6bb7941cf34fcc02b6fc4163a80e350b5bfd4c47R420)

collectors/carbonblack/collector.js
rakhimundhada15: Parsing the json  in al-aws-collector [checkin function ](https://github.com/alertlogic/al-aws-collector-js/pull/62/files#diff-c06f5f0e08865e628912333b6bb7941cf34fcc02b6fc4163a80e350b5bfd4c47R411)

collectors/gsuite/local/events/event_poll.json
```diff
@@ -3,7 +3,7 @@
         {
             "messageId": "f67a45fc-ab43-4df1-9f2f-585bb2043122",
             "receiptHandle": "AQEB78LmKaaZ+uj8DVnc/O2zQcAe+qKSi3ZGTSBFssIRSmpwwzUEi2vhPTaIBnhOoh0aoRxVjWdoXO3ZloINfMxmcjycP3KC0WXwcWokoOc3iMCdqhYg0NcOhQW1X0ixc79C9/5/XF1xGd79vLhFL7KvRjjiT4sOaSxlAv6v2fJ5eDETnp7CRa5pocCF4EO2su0M4/TnlLreGfsY+C+/tH+r19AM+d3Jt5dbNMrKMWRRZ7/PTjczkIM7U38AHPuusuBz9uzA5yMQGOMI8FPXfqgcafEN17JqKuNSd/l54v1+s9rBQSzL/MH9wZ0XpZditcpe5pTc+dGuRHOXbFK2A0YUQCvRq1Ed8tfr68uELTQK5jWHH/zPnEMWsTyco9VuNCKr",
-            "body": "{\n  \"priv_collector_state\": {\n    \"since\": \"2019-11-19T10:00:00Z\",\n    \"until\": \"2020-01-24T10:30:00Z\",\n  \"nextPage\": null,\n   \"application\": \"login\",\n   \"poll_interval_sec\": 900,\n \"apiQuotaResetDate\": \"2020-02-13T08:00:59.691Z\"  }\n}",
+            "body": "{\n  \"priv_collector_state\": {\n    \"since\": \"2019-11-19T10T10:00:00Z\",\n    \"until\": \"2020-01-24T10:30:00Z\",\n  \"nextPage\": null,\n   \"stream\": \"login\",\n   \"poll_interval_sec\": 900,\n \"apiQuotaResetDate\": \"2020-02-13T08:00:59.691Z\"  }\n}",
```
OHaganA, line 6: Duplicate `T10` in the timestamp here

daniel-greening (changes requested):
Unfortunately we can't just change the collector state like this, as I mentioned here: https://github.com/alertlogic/paws-collector/pull/171/files#diff-0abfd24bb844ebded421f7a98ba7e172f6c4242808fc12f69baefe73528f8e41R39-R40 we need to consider existing collectors before modifying private collector state, otherwise we'll break them when they update to the latest version.

daniel-greening:
Carbon Black implementation looks good

collectors/gsuite/local/events/event_poll.json
rakhimundhada15: Sure.I will update

daniel-greening (approved):

PR #160, opened 2020-11-23
ENG-18709 : Handle the additional collector metrics for client errors in each collector

rakhimundhada15:
1. Added errorCode if not available in error object to showcase client error on DDMetric
2. Added unit test case to check error

### DIFF
--- a/collectors/gsuite/collector.js
+++ b/collectors/gsuite/collector.js
@@ -124,6 +124,10 @@ class GsuiteCollector extends PawsCollector {
                         });
                     }
                     else {
+                        // set errorCode if not available in error object to showcase client error on DDMetrics
+                        if (error.errors && error.errors.length > 0 && error.errors[0].reason) {
+                            error.errorCode = error.errors[0].reason;
+                        };
                         return callback(error);
                     }
 
--- a/collectors/o365/o365_collector.js
+++ b/collectors/o365/o365_collector.js
@@ -164,10 +164,15 @@ class O365Collector extends PawsCollector {
 
             return callback(null, logs, newState, newState.poll_interval_sec);
         }).catch(err => {
-            if(err.message){
+            if (err.message) {
                 let message = err.message;
-                let error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
-                err.errorCode = error.error;
+                // set errorCode if not available in error object to showcase client error on DDMetric
+                try {
+                    let error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
+                    err.errorCode = error.error;
+                } catch (exception) {
+                    return callback(err);
+                }
             }
             console.error(`O365000003 Error in collection: ${err.message}`);
             return callback(err);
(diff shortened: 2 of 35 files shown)


collectors/gsuite/collector.js
```diff
@@ -124,6 +124,10 @@ class GsuiteCollector extends PawsCollector {
                         });
                     }
                     else {
+                        // set errorCode if not available in error object to showcase client error on DDMetrics
+                        if (error.errors && error.errors.length > 0 && error.errors[0].reason) {
+                            error.errorCode = error.errors[0].reason;
```
JDragovichAlertLogic, line 129: Is that going to be short enough to work as an error code?

collectors/o365/o365_collector.js
```diff
@@ -164,6 +164,11 @@ class O365Collector extends PawsCollector {
 
             return callback(null, logs, newState, newState.poll_interval_sec);
         }).catch(err => {
+            if(err.message){
+                let message = err.message;
+                let error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
```
JDragovichAlertLogic, line 169: I'm sure youve got a good reason for doing this, but that looks fragile to me. And considering its an MS api, it might change with little warning.

I would wrap this line in a try/catch block and return some sort of generic error if that parsing fails

JDragovichAlertLogic (changes requested):
Thank you Rakhi. A few comments

collectors/gsuite/collector.js
rakhimundhada15: yes, It is short and also explain error.  The reason we are getting is like 'insufficientPermissions' . We are also getting code (ex.403). Shall I use code ?

collectors/o365/o365_collector.js
rakhimundhada15: wrap the paring error and pushed the changes

collectors/gsuite/collector.js
rakhimundhada15: As the reason field is  short and more granular than the http code and we had already used reason field  for checking throttling error as well. To keep code consistent, using reason field as error code.

collectors/o365/o365_collector.js
```diff
@@ -164,6 +164,15 @@ class O365Collector extends PawsCollector {
 
             return callback(null, logs, newState, newState.poll_interval_sec);
         }).catch(err => {
+            if (err.message) {
+                let message = err.message;
+                try {
+                    let error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
+                    err.errorCode = error.error;
+                } catch (exception) {
```
JDragovichAlertLogic, line 172: I wouldnt return the parsing exception because that will mask the actual collector error.

This try/catch block is more for handling parsing errors gracefully. I would return the `err` object in the callback.

JDragovichAlertLogic (changes requested):
Thanks Rakhi. One final comment.

collectors/o365/o365_collector.js
rakhimundhada15: okay i will update.

JDragovichAlertLogic (approved):
Looks good. thanks Rakhi

PR #161, opened 2020-11-24
ENG-0000 : Updated the required properties in sam-template.yaml

rakhimundhada15:
Updated the required properties in sam template  and  env.json.tmpl to resolved errors while running  in local.


kkuzmin:
@rakhimundhada15 please rebase and let's merge it

PR #163, opened 2020-12-08
change themis templates for new themis params

JDragovichAlertLogic:
### Problem Description
TH ethemis templates done work because the themis api changed

### Solution Description
change the templates to conform to the new interface


OHaganA (approved):
LGTM, is it just Sophos and Sophos SIEM that's broken?

JDragovichAlertLogic:
> LGTM, is it just Sophos and Sophos SIEM that's broken?

yeah from what i can see. those are the only ones with validation templates

PR #164, opened 2020-12-09
ENG-19900 : Additional collector metrics for ingest error

rakhimundhada15:


paws_collector.js
```diff
@@ -479,7 +491,7 @@ class PawsCollector extends AlAwsCollector {
             ],
             Namespace: 'PawsCollectors'
         };
-        let errorCode = typeof (error) === 'object' && error.errorCode ? error.errorCode : 'unknown';
+        let errorCode = typeof (error) === 'object' && error.errorCode ? error.errorCode : `unknown_${this.collector_id}`;
```
JDragovichAlertLogic, line 494: @kkuzmin what do you think of adding the collector id to the "unknown" error tag? Are we going to get too many tags? I think it should be ok.

paws_collector.js
```diff
@@ -450,6 +454,14 @@ class PawsCollector extends AlAwsCollector {
         this.reportDDMetric('api_throttling', 1)
         return cloudwatch.putMetricData(params, callback);
     };
+    /**
+     * Report the error to Ingest api service and show case on DDMetrics
+     * @param error 
+     */
+    reportErrorToIngestApi(error) {
```
JDragovichAlertLogic, line 461: Not sure how this is reporting the error to ingest. Are we handling this through a collector status? 

Alos shouldn't this have some sort of callback? arent we going to have a race condition with the async callback?

like this https://github.com/alertlogic/paws-collector/blob/6a6dcc561aa2bee5fa9085ed4ab3016637ba8c89/paws_collector.js#L387-L389

paws_collector.js
```diff
@@ -392,8 +392,12 @@ class PawsCollector extends AlAwsCollector {
             },
             function(logs, privCollectorState, nextInvocationTimeout, asyncCallback) {
                 console.info('PAWS000200 Log events received ', logs.length);
-                return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null, function(err) {
-                    return asyncCallback(err, privCollectorState, nextInvocationTimeout);
+                return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null, (err) => {
+                    if (err) {
+                        collector.reportErrorToIngestApi(err);
```
JDragovichAlertLogic, line 397: see comment below. I think we need to add a callback to this function

JDragovichAlertLogic (changes requested):
Thanks for this Rakhi. a few comments

paws_collector.js
daniel-greening: I've been advised against doing this in the past. Typically you want to keep instance specific dimensions out of datadog metrics, apparently due to cost issues, though maybe this has changed. 

I'll reach out to cloud ops team about it.

kkuzmin: I haven't heard anything changed about tags. I advise to not do this.
If required we should always be able to find collector ID in service logs in kibana

JDragovichAlertLogic: Sounds reasonable.

The one issue with these client errors is that they never reach our back end. So if we see client errors in DD, we don't really have a good way of finding out which collector is causing the error.

crazy idea. could we put all the cloudwatch logs for paws collectors in search?

JDragovichAlertLogic (changes requested):

daniel-greening (approved):

daniel-greening (approved):

PR #167, opened 2020-12-18
ENG-0000: Removed the collector type from dd metrics name

rakhimundhada15:
Removed the **collector type** from datadog metrics name and used  the collector type as the tag (`paws_platform:${this.pawsCollectorType}`) while creating graph on DD.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alertlogic/paws-collector",
-  "version": "2.0.3",
+  "version": "2.0.4",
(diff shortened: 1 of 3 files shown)


daniel-greening (approved):

OHaganA (approved):

kkuzmin:
@rakhimundhada15 Not sure why we need this change?

kkuzmin:
I think we don't need it.

PR #168, opened 2020-12-31
Add health check to see if there are any updates states in the last 3…

JDragovichAlertLogic:
…0 mins.

### Problem Description
Right now we don't have any health checks to see if the collector is running. 

### Solution Description
this checks to see if there are any states updated by the collector in the last 30 mins. It does this by calling hte DDB dedupe table.

This should help us flag up dead collectors.

### DIFF
--- a/healthcheck.js
+++ b/healthcheck.js
@@ -1,0 +1,39 @@
+const AWS = require('aws-sdk');
+const moment = require('moment');
+
+function checkLastState(callback){
+    const collector = this;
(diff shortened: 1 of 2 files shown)


kkuzmin:
I don't think this belong to collector itself. I'd suggest to extract it into some ops script we we can run from time to time.

JDragovichAlertLogic:
Thats fair. I guess the advantage to this approach is that it should create a remediation in the health console. So customers can self serve, or at least alert support if the collector is broken.

PR #171, opened 2021-01-27
add proposal for new state ddb

daniel-greening:
### Problem Description
- due to consistent errors, streams may stop collecting permanently due to lost SQS messages

### Solution Description
- Introduce new state DDB that tracks collection state for each stream of a collector and new health check

PR #172, opened 2021-01-28
ENG-17586 :  [paws-collectors] Stream specific status reporting 

rakhimundhada15:
1.Passed the streams array to al-aws-collector to post ok status in checkin event
2.Send the stream from done function to post the error  status specific to stream.
3. Added collectorType parameter in prepareErrorStatus to override the prepareErrorStatus of al-aws-collector when there is stream specific error status.

Prerequisite:
[al-aws-collector/62](https://github.com/alertlogic/al-aws-collector-js/pull/62)

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -18,6 +18,7 @@ const moment = require('moment');
 const ddLambda = require('datadog-lambda-js');
 
 const AlAwsCollector = require('@alertlogic/al-aws-collector-js').AlAwsCollector;
+const AlAwsUtil = require('@alertlogic/al-aws-collector-js').Util;
 const packageJson = require('./package.json');
 
 const CREDS_FILE_PATH = '/tmp/paws_creds';
@@ -657,6 +658,20 @@ class PawsCollector extends AlAwsCollector {
         // Current state message will be removed by Lambda trigger upon successful completion
         sqs.sendMessage(params, callback);
     };
+
+    /**
+     * This function to set collector_streams in environment variable for exsisting collector
+     * Streams is send to AL-aws-collector to post stream specific status if there is no error.
+     * @param {*} streams 
+     */
+    setCollectorStreamsEnv(streams) {
+        let collectorStreams = { collector_streams: streams };
+        return AlAwsUtil.setEnv(collectorStreams, (err) => {
+            if (err) {
+                console.error('Paws error while adding collector_streams in environment variable')
+            }
+        });
+    }
 
     /**
      * @function collector callback to initialize collection state
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -96,14 +96,14 @@ class PawsCollector extends AlAwsCollector {
         }
     }
 
-    constructor(context, {aimsCreds, pawsCreds}, childVersion, healthChecks = [], statsChecks = []) {
+    constructor(context, {aimsCreds, pawsCreds}, childVersion,streams =[],healthChecks = [], statsChecks = []) {
         const version = childVersion ? childVersion : packageJson.version;
         const endpointDomain = process.env.paws_endpoint.replace(DOMAIN_REGEXP, '');
         super(context, 'paws',
               AlAwsCollector.IngestTypes.LOGMSGS,
               version,
               aimsCreds,
-              null, healthChecks, statsChecks);
+              null, healthChecks, statsChecks,streams);
         console.info('PAWS000100 Loading collector', process.env.paws_type_name);
         this._pawsCreds = pawsCreds;
         this._pawsCollectorType = process.env.paws_type_name;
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -170,8 +170,8 @@ class PawsCollector extends AlAwsCollector {
         return Object.assign(pawsProps, baseProps);
     };
 
-    prepareErrorStatus(errorString, streamName = 'none') {
-        return super.prepareErrorStatus(errorString, streamName, this.pawsCollectorType);
+    prepareErrorStatus(errorString, streamName = 'none', collectorType = this.pawsCollectorType) {
```
kkuzmin, line 173: why do we need 3rd parameter here? Is it actually used in any collector?

paws_collector.js
```diff
@@ -96,14 +96,14 @@ class PawsCollector extends AlAwsCollector {
         }
     }
 
-    constructor(context, {aimsCreds, pawsCreds}, childVersion, healthChecks = [], statsChecks = []) {
+    constructor(context, {aimsCreds, pawsCreds}, childVersion,streams =[],healthChecks = [], statsChecks = []) {
```
kkuzmin, line 99: @daniel-greening has a proposal to use `collector_stream` env var. Please see [here](https://github.com/alertlogic/paws-collector/pull/171#pullrequestreview-578447101).
I think it is worth implementing it right away.
@daniel-greening @rakhimundhada15 Thoughts?

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: When we call prepareErrorStatus from al-aws-collector [done function](https://github.com/alertlogic/al-aws-collector-js/pull/62/files#diff-c06f5f0e08865e628912333b6bb7941cf34fcc02b6fc4163a80e350b5bfd4c47R188) it actually call Paws collector(above )method and 3rd parameter is not match so  I am not getting collectorType, So I have updated this method.

paws_collector.js
rakhimundhada15: I can implement and As  per my understanding we have to made changes in below files of paws-collector:
[ catalog.json](https://algithub.pd.alertlogic.net/defender/applications/blob/master/priv/catalog.json#L1514)
[salesforce-collector.template](https://github.com/alertlogic/paws-collector/blob/master/collectors/salesforce/cfn/salesforce-collector.template#L113)
[collector.js](https://github.com/alertlogic/paws-collector/blob/master/collectors/salesforce/collector.js#L39) 
[sam-template](https://github.com/alertlogic/paws-collector/blob/master/collectors/salesforce/local/sam-template.yaml#L30) 
Then we don't have to pass stream array from each collector as we can use common variable in Paws-collector.js

rakhimundhada15: As discussed with Dan , I am directly using collector_stream here and rest of change in other file he will take care.

paws_collector.js
```diff
@@ -96,14 +96,15 @@ class PawsCollector extends AlAwsCollector {
...
+    constructor(context, {aimsCreds, pawsCreds}, childVersion,healthChecks = [], statsChecks = []) {
         const version = childVersion ? childVersion : packageJson.version;
         const endpointDomain = process.env.paws_endpoint.replace(DOMAIN_REGEXP, '');
+        let collectorStreams = process.env.collector_stream ? JSON.parse(process.env.collector_stream) : [];
         super(context, 'paws',
               AlAwsCollector.IngestTypes.LOGMSGS,
               version,
               aimsCreds,
-              null, healthChecks, statsChecks);
+              null, healthChecks, statsChecks,collectorStreams);
```
kkuzmin, line 107: please add a space here.

paws_collector.js
```diff
@@ -96,14 +96,15 @@ class PawsCollector extends AlAwsCollector {
         }
     }
 
-    constructor(context, {aimsCreds, pawsCreds}, childVersion, healthChecks = [], statsChecks = []) {
+    constructor(context, {aimsCreds, pawsCreds}, childVersion,healthChecks = [], statsChecks = []) {
         const version = childVersion ? childVersion : packageJson.version;
         const endpointDomain = process.env.paws_endpoint.replace(DOMAIN_REGEXP, '');
+        let collectorStreams = process.env.collector_stream ? JSON.parse(process.env.collector_stream) : [];
```
kkuzmin, line 102: please add one more extra check to make sure `JSON.parse(process.env.collector_stream)` is an array

kkuzmin (changes requested):

OHaganA (approved):
LGTM

daniel-greening (approved):
Looks good

PR #173, opened 2021-02-17
convert collection_start_ts to ISO8601

daniel-greening:
### Problem Description
- cases where customers have used valid date times such as `2021-02-16T21:46:54Z` (missing millisecond time)
- auth0 throws error when attempting to call apis without milliseconds timestamp

### Solution Description
- convert such datetime strings into valid ISO8601 format

### DIFF
--- a/collectors/auth0/utils.js
+++ b/collectors/auth0/utils.js
@@ -1,7 +1,9 @@
+const moment = require('moment');
+
 function getAPILogs(auth0Client, state, accumulator, maxPagesPerInvocation) {
     let pageCount = 0;
 
-    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date=[" + state.since + " TO *]", per_page:100, sort: "date:1" };
+    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date=[" + moment(state.since).toISOString() + " TO *]", per_page:100, sort: "date:1" };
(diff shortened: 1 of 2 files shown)


OHaganA (approved):

collectors/auth0/utils.js
```diff
@@ -1,7 +1,9 @@
+const moment = require('moment');
+
 function getAPILogs(auth0Client, state, accumulator, maxPagesPerInvocation) {
     let pageCount = 0;
 
-    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date=[" + state.since + " TO *]", per_page:100, sort: "date:1" };
+    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date=[" + moment(state.since).toISOString() + " TO *]", per_page:100, sort: "date:1" };
```
scott-jones-AlertLogic, line 6: Should the query use `date:[...` instead of `date=[...`, like this?

```suggestion
    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date:[" + moment(state.since).toISOString() + " TO *]", per_page:100, sort: "date:1" };
```

From Auth0 docs (https://auth0.com/docs/logs/log-search-query-syntax#example-queries):
Search for all logs from December 10, 2018 forward | date:[2018-12-10 TO *]
-- | --

PR #180, opened 2021-03-15
VTT-2603: Create SSM secure parameter by Lambda function

rakhimundhada15:
### Problem Description
The size of the key is grater than 4kb , when encrypted and encoded as base64, is too large to be stored in the CustomResource before being stored in SSM .

### Solution Description

Instead  encrypting the paws secret using EncryptLambdaFunction, we should store the secret directly in **SSM Parameter Store as a SecureString**. Create SSM  SecureString parameter with the help of Lambda function instate of CFN template (as 
 mentioned on AWS doc - `SecureString is not currently supported for AWS CloudFormation templates`.) .

### Implementation
1. Added Lambda Function (CreateSsmParameterLambdaFunction ) which will create the SSM parameter with Secure String and return the Parameter name and version.If `event.RequestType == 'Delete'` then Delete the SSM parameter.

2. Modified the  EncryptLambdaPolicy to allow (ssm:PutParameter) which required to Create SSM parameter by Lambda function and  updated the resources to check logs of  CreateSsmParameterLambdaFunction.

3. Added `ssm_direct` in environment variable and set to "true" for new deployed collect. This ssm_direct  variable use in the paws-collector code to check whether we should 1) fetch the secret from SSM, base64 decode it, then decrypt using KMS (current way of getting paws secret); or 2) request the decrypted parameter directly from SSM when ssm_direct is set.

4. Removed PawsSecretParam( Type :AWS::SSM::Parameter) and now this has been created by lambda function.

### DIFF
--- a/cfn/paws-collector.template
+++ b/cfn/paws-collector.template
@@ -563,11 +563,27 @@
                                 "    });\n",
                                 "}\n",
                                 "\n",
+                                "function deleteSSMParam(event, context) {\n",
+                                "    var params = {\n",
+                                "        Name: event.ResourceProperties.Name\n",
+                                "    };\n",
+                                "    const ssm = new AWS.SSM();\n",
+                                "    ssm.deleteParameter(params, function(err, data) {\n",
+                                "        if (err) {\n",
+                                "            console.log(err, err.stack); // an error occurred\n",
+                                "            return response.send(event, context, response.FAILED);\n",
+                                "        }\n",
+                                "        return response.send(event, context, response.SUCCESS, data);\n",
+                                "    });\n",
+                                "}\n",
                                 "\n",
                                 "exports.handler = (event, context, callback) => {\n",
                                 "    if (event.ResourceType == 'AWS::CloudFormation::CustomResource' &&\n",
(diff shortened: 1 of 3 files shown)


cfn/paws-collector.template
```diff
@@ -514,14 +651,21 @@
...
+                    "Ref": "PawsSecret"
+                },
+                "Tier": {
+                    "Ref": "PawsSecretParamTier"
+                },
+                "Tags": [{
+                        "Key": "StackId",
+                        "Value": {
+                            "Ref": "AWS::StackId"
+                        }
```
scott-jones-AlertLogic, line 667: cosmetic: could you please format this using standard JSON formatting? e.g.
```json
{
    "Tags": [
        {
            "Key": "StackId",
            "Value": {
                "Ref": "AWS::StackId"
            }
        }
    ]
}
```

scott-jones-AlertLogic (approved):
+1, just a small comment about indentation in the cfn template

Nice work; thank you for handling this!

cfn/paws-collector.template
rakhimundhada15: sure

surajkadam (approved):

surajkadam (approved):

PR #193, opened 2021-07-08
Handle the MS management api issues

rakhimundhada15:
### Problem Description
Getting the MS management api below errors:

1. Maximum payload size exceeded
2. Content older than 7 days cannot be retrieved ;when we make call for each ContentUri in ayncPool. 
ex : `Content requested with the key 20210630122427504000475$20210630122427504000475$audit_sharepoint$Audit_SharePoint$na0047 has already expired. Content older than 7 days cannot be retrieved`.

### Solution Description

1. Reduce the time duration by 2 ;If duration is 1hr then we set to state.until to 30min.
2. Change the PARTIAL_WEEK to PARTIAL_WEEK_HOURS and set the 6 days 23hr to avoid the error `Content older than 7 while making call to each contentUri.

### DIFF
--- a/collectors/o365/o365_collector.js
+++ b/collectors/o365/o365_collector.js
@@ -177,11 +177,16 @@ class O365Collector extends PawsCollector {
             if (newState) {
                 return callback(null, [], newState, newState.poll_interval_sec);
             }
-            
-            this._setUserUnderstandableErrorMessage(err, callback);
-           
-        })
-
+ 
+            if (!err.code && err.message) {
+              const formatedError = this._formatErrorMessage(err);
+              console.error(`O365000003 Error in collection: ${err.message}`);
+              return callback(formatedError);
+            } else {
+                // if error is string or don't have err.code  in error object, then return complete error object/string.
+                return callback(err);
+            }
+        });
     }
 
     _getNextCollectionState(curState) {
@@ -268,30 +273,32 @@ class O365Collector extends PawsCollector {
         return null;
     }
 
-    _setUserUnderstandableErrorMessage(err, callback){
-        if (!err.code && err.message) {
-            let message = err.message;
-            // set errorCode if not available in error object to showcase client error on DDMetric
-            try {
-                let error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
-                err.errorCode = error.error ? error.error : error.error_codes[0];
-                if (error.error_codes) {
-                    if (error.error_codes[0] === 7000215) {
-                        return callback("Error code [7000215]. Invalid client secret is provided.");
-                    }
-                    if (error.error_codes[0] === 700016) {
-                        return callback("Error code [700016]. Invalid client ID is provided.");
-                    }
-                    if (error.error_codes[0] === 90002) {
-                        return callback("Error code [90002]. Please make sure you have the correct tenant ID or this may happen if there are no active subscriptions for the tenant. Check with your subscription administrator.");
-                    }
+    /**
+     * Format the error message to user understandable message
+     * @param {*} err 
+     * @returns 
+     */
+    _formatErrorMessage(err) {
+        const message = err.message;
+        // set errorCode if not available in error object to showcase client error on DDMetric
+        try {
+            const error = JSON.parse(message.slice(message.indexOf('{'), message.lastIndexOf('}') + 1));
+            err.errorCode = error.error ? error.error : error.error_codes[0];
+            if (error.error_codes) {
+                if (error.error_codes[0] === 7000215) {
+                    err.message = 'Error code [7000215]. Invalid client secret is provided.';
+                }
+                if (error.error_codes[0] === 700016) {
+                    err.message = 'Error code [700016]. Invalid client ID is provided.';
+                }
+                if (error.error_codes[0] === 90002) {
+                    err.message = 'Error code [90002]. Please make sure you have the correct tenant ID or this may happen if there are no active subscriptions for the tenant. Check with your subscription administrator.';
                 } 
-            } catch (exception) {
-                return callback(err);
             }
+            return err;
+        } catch (exception) {
+            return err;
         }
-        console.error(`O365000003 Error in collection: ${err.message}`);
-        return callback(err);
     }
 }
 
--- a/collectors/o365/o365_collector.js
+++ b/collectors/o365/o365_collector.js
@@ -173,14 +173,9 @@ class O365Collector extends PawsCollector {
             if (typeof err === 'object') {
                 err.errorCode = err.code ? err.code : (err.statusCode ? err.statusCode : err.status);
             }
-            // Handle the MS management api maximum palyload issue
-            if (err.message && err.message.indexOf('Maximum payload size exceeded') !== -1) {
-                let min = moment(state.until).diff(state.since, 'minutes');
-                if (min > 1) {
-                    state.until = moment(state.since).add(min / 2, 'minutes').toISOString();
-                }
-                console.warn(`Reduce the time duration by ${min / 2} min due to Maximum payload issue`);
-                return callback(null, [], state, state.poll_interval_sec);
+            let newState = this._handleMSManagementApiError(err, state);
+            if (newState) {
+                return callback(null, [], newState, newState.poll_interval_sec);
             }
             if (!err.code && err.message) {
                 let message = err.message;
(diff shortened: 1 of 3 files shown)


collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,20 @@ class O365Collector extends PawsCollector {
 
             return callback(null, logs, newState, newState.poll_interval_sec);
         }).catch(err => {
-            if (err.message) {
+            // set errorCode to showcase client error on DDMetric
+            if (typeof err === 'object') {
+                err.errorCode = err.code ? err.code : (err.statusCode ? err.statusCode : err.status);
+            }
+            // Handle the MS management api maximum palyload issue
+            if (err.message && err.message.indexOf('Maximum payload size exceeded') !== -1) {
```
kkuzmin, line 177: can we reply on some `statusCode` or similar instead of comparing strings?

collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,20 @@ class O365Collector extends PawsCollector {
...
+                let min = moment(state.until).diff(state.since, 'minutes');
+                if (min > 1) {
+                    state.until = moment(state.since).add(min / 2, 'minutes').toISOString();
+                }
+                console.warn(`Reduce the time duration by ${min / 2} min due to Maximum payload issue`);
```
kkuzmin, line 182: please output new `since` and `until` values in the message.

collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,20 @@ class O365Collector extends PawsCollector {
...
+                return callback(null, [], state, state.poll_interval_sec);
```
kkuzmin, line 183: I think we need to change `state.poll_interval_sec` to `1` second in order to speed up collection

collectors/o365/test/o365_test.js
```diff
@@ -383,7 +383,7 @@ describe('O365 Collector Tests', function() {
                     const callArgs = subscriptionsContentStub.getCall(0).args;
 
                     assert.equal(callArgs[0], curState.stream);
-                    assert.equal(moment().diff(callArgs[1], 'days'), 7);
+                    assert.equal(moment().diff(callArgs[1], 'days'), 6); // difference is 6 days 23hr
```
kkuzmin, line 386: please check hours diff instead, same below.

collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,20 @@ class O365Collector extends PawsCollector {
 
             return callback(null, logs, newState, newState.poll_interval_sec);
         }).catch(err => {
-            if (err.message) {
+            // set errorCode to showcase client error on DDMetric
```
kkuzmin, line 172: could you please extract the new code below into a separate function, like `_handleMaxPayloadExceededError`

collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,20 @@ class O365Collector extends PawsCollector {
...
+            // Handle the MS management api maximum palyload issue
+            if (err.message && err.message.indexOf('Maximum payload size exceeded') !== -1) {
+                let min = moment(state.until).diff(state.since, 'minutes');
+                if (min > 1) {
+                    state.until = moment(state.since).add(min / 2, 'minutes').toISOString();
+                }
+                console.warn(`Reduce the time duration by ${min / 2} min due to Maximum payload issue`);
+                return callback(null, [], state, state.poll_interval_sec);
+            }
+            if (!err.code && err.message) {
```
kkuzmin, line 185: could you please also add a handler for expired content. I think we currently just fail the invocation instead of just acknowledging this error and proceeding with collection. We should probably handle this error in https://github.com/alertlogic/paws-collector/pull/193/files#diff-458b6045c8dfaadace71e6dbd57f1b5625d5c3f985af520c6b352b5111cff86cR150

kkuzmin (changes requested):

collectors/o365/o365_collector.js
rakhimundhada15: Actually i don't get complete error object. We only saw error message on cloud watch and it also not reproduce on local to get error object; so i am checking from message.

kkuzmin (changes requested):
@rakhimundhada15 looks good overall but I think we can do better refactoring error handling.

collectors/o365/o365_collector.js
```diff
@@ -241,6 +250,44 @@ class O365Collector extends PawsCollector {
...
+                state.poll_interval_sec = 1;
+            }
+            console.warn(`Collecting data from ${state.since} to ${state.until} to handle Maximum payload issue`);
+            return state;
+        }
+
+        if (err.response && err.response.body) {
+            let responseBody = JSON.parse(err.response.body);
+            if (responseBody.error.code === 'AF20051') {
+                let min = moment(state.until).diff(state.since, 'minutes');
```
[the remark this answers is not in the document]
rakhimundhada15: updating with currentInterval .

collectors/o365/o365_collector.js
```diff
@@ -168,7 +169,15 @@ class O365Collector extends PawsCollector {
...
-            if (err.message) {
+            // set errorCode to showcase client error on DDMetric
+            if (typeof err === 'object') {
+                err.errorCode = err.code ? err.code : (err.statusCode ? err.statusCode : err.status);
+            }
+            let newState = this._handleMSManagementApiError(err, state);
+            if (newState) {
+                return callback(null, [], newState, newState.poll_interval_sec);
+            }
+            if (!err.code && err.message) {
```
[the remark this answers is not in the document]
rakhimundhada15: Here we are just setting user understandable message ,so i will create new method _setUserUndestandlableErrorMessage and move the code.

collectors/o365/o365_collector.js
```diff
@@ -241,6 +250,44 @@ class O365Collector extends PawsCollector {
...
+    /**
+     * 1.Handle the Maxpayload exceeded error;reduce the pull duration by 2 and set new time in state.untill
+     * 2.Expired content error can be occur when start date is older than 7day and we reset the date to last 7days
+     * If data for 1st 1hr is huge and within time it will not able to pull the data from array of contentUri using asyncPool;
+     * Then set the satrt time to 15 min later,in this case we loss the data for 15 min but will resolve the issue and get remaining data without any error
+     * @param {*} err 
+     * @param {*} state
+     * @returns 
+     */
+    _handleMSManagementApiError(err, state) {
```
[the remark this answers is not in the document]
rakhimundhada15: Initially i have also spited in two different function  but latter i have removed as both returning the new state and we are handling the MS api errors only in both , so keep single method.

collectors/o365/o365_collector.js
```diff
@@ -241,6 +250,44 @@ class O365Collector extends PawsCollector {
...
+            }
+            console.warn(`Collecting data from ${state.since} to ${state.until} to handle Maximum payload issue`);
+            return state;
+        }
+
+        if (err.response && err.response.body) {
+            let responseBody = JSON.parse(err.response.body);
+            if (responseBody.error.code === 'AF20051') {
+                let min = moment(state.until).diff(state.since, 'minutes');
+                if (min >= 1) {
```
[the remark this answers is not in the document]
rakhimundhada15: sure , i will remove the condition.

collectors/o365/o365_collector.js
```diff
@@ -241,6 +250,44 @@ class O365Collector extends PawsCollector {
...
+                state.until = moment(state.since).add(min / 2, 'minutes').toISOString();
+                state.poll_interval_sec = 1;
+            }
+            console.warn(`Collecting data from ${state.since} to ${state.until} to handle Maximum payload issue`);
+            return state;
+        }
+
+        if (err.response && err.response.body) {
+            let responseBody = JSON.parse(err.response.body);
+            if (responseBody.error.code === 'AF20051') {
```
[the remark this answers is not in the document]
kkuzmin: Please use constant here.

collectors/o365/o365_collector.js
```diff
@@ -168,29 +169,17 @@ class O365Collector extends PawsCollector {
...
+                err.errorCode = err.code ? err.code : (err.statusCode ? err.statusCode : err.status);
             }
-            console.error(`O365000003 Error in collection: ${err.message}`);
-            return callback(err);
+            let newState = this._handleMSManagementApiError(err, state);
+            if (newState) {
+                return callback(null, [], newState, newState.poll_interval_sec);
+            }
+            
+            this._setUserUnderstandableErrorMessage(err, callback);
```
kkuzmin, line 181: this function is not only formats messages. ANd I'd prefer to keep final callback in here.  Let's replace/rename it with
```
if (!err.code && err.message) 
    const msg = this._formatErrorMessage(err);
    console.error(`O365000003 Error in collection: ${err.message}`);
    return callback(msg)
else 
 return callback(null)
```

kkuzmin (changes requested):

collectors/o365/o365_collector.js
rakhimundhada15: sure, I will rename the function name to _formatErrorMessage.

kkuzmin (approved):

PR #195, opened 2021-07-19
Changes to handle the Max payload error

rakhimundhada15:
### Problem Description
For O365 collector when we pulling the data for 1hr duration, if data size > 10Mb then al-collector.js  [buildPayload](https://github.com/alertlogic/al-collector-js/blob/master/al_log.js#L97) method return Max payload issue.

### Solution Description
Reduce the pull interval by 2 and accordingly updated the state value

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -487,45 +487,33 @@ class PawsCollector extends AlAwsCollector {
         });
     }
 
-    batchLogProcess(logs, privCollectorState, nextInvocationTimeout, asyncCallback) {
+    batchLogProcess(logs, privCollectorState, nextInvocationTimeout, callback) {
         let collector = this;
-        return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null, (err) => {
+        return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null,  (err) => {
            
             if (err) {
-                // MS api return the logs but if build payload size > 10MB, our buildPayload method return Max payload issue. So we split logs array the in two part and call the processLog again.
-               
+                /**
+                 * MS api return the logs but if build payload size > 10MB, and split logs array the in two part.
+                 * Call the processLog again to send Logmsgs to ingest. 
+                 */
                 if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
-                    let batchLogs = [];
-                    let halfLogs = [];
-                    let loglength = logs.length / 2;
-                    for (const index in logs) {
-                        console.log(index,logs[index]);
-                        if (index < loglength) {
-                            halfLogs.push(logs[index]);
-                            if (parseInt(index) === (loglength - 1)) {
-                                batchLogs.push(halfLogs);
-                            }
-                        }
-                        else {
-                            halfLogs = [];
-                            loglength = logs.length;
-                            halfLogs.push(logs[index]);
-                            if (parseInt(index) == logs.length-1) {
-                                batchLogs.push(halfLogs);
-                            }
-                        }
-                    }
-                    batchLogs.map(batchLog => {
-                         collector.batchLogProcess(batchLog, privCollectorState, nextInvocationTimeout, asyncCallback);
+
+                    let half = logs.length / 2;
+                    const indexArray = [
+                        { start: 0, stop: half },
+                        { start: half, stop: logs.length }
+                    ];
+
+                    async.reduce(indexArray, 0, (memo, logpart, asyncCallback) => {
+                        collector.batchLogProcess(logs.slice(logpart.start, logpart.stop), privCollectorState, nextInvocationTimeout, asyncCallback);
                     });
-                   
                 } else {
                     collector.reportErrorToIngestApi(err, () => {
-                        return asyncCallback(err);
+                        return callback(err);
                     });
                 }
             } else {
-                return asyncCallback(null, privCollectorState, nextInvocationTimeout);
+                return callback(null, privCollectorState, nextInvocationTimeout);
             }
         });
     }
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -516,8 +516,8 @@ class PawsCollector extends AlAwsCollector {
                             });
                         });
                     });
-                    const results = await Promise.all(promises);
-                    return callback(null, results[results.length-1], nextInvocationTimeout);
+                    await Promise.all(promises);
+                    return callback(null, privCollectorState, nextInvocationTimeout);
                 } else {
                     collector.reportErrorToIngestApi(err, () => {
                         return callback(err);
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -461,6 +439,49 @@ class PawsCollector extends AlAwsCollector {
         });
     };
 
+    batchLogProcess(logs, privCollectorState, nextInvocationTimeout, asyncCallback) {
+        let collector = this;
+        return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null, (err) => {
+           
+            if (err) {
+                // MS api return the logs but if build payload size > 10MB, our buildPayload method return Max payload issue. So we split logs array the in two part and call the processLog again.
```
kkuzmin, line 447: please make it a multiline comment

paws_collector.js
```diff
@@ -461,6 +439,49 @@ class PawsCollector extends AlAwsCollector {
...
+               
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
+                    let batchLogs = [];
+                    let halfLogs = [];
+                    let loglength = logs.length / 2;
+                    for (const index in logs) {
+                        console.log(index,logs[index]);
```
kkuzmin, line 454: please remove debug logs.

paws_collector.js
```diff
@@ -461,6 +439,49 @@ class PawsCollector extends AlAwsCollector {
         });
     };
 
+    batchLogProcess(logs, privCollectorState, nextInvocationTimeout, asyncCallback) {
```
kkuzmin, line 442: please rename `asyncCallback ` to `callback`

paws_collector.js
```diff
@@ -461,6 +439,49 @@ class PawsCollector extends AlAwsCollector {
...
+        return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null, (err) => {
+           
+            if (err) {
+                // MS api return the logs but if build payload size > 10MB, our buildPayload method return Max payload issue. So we split logs array the in two part and call the processLog again.
+               
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
+                    let batchLogs = [];
+                    let halfLogs = [];
+                    let loglength = logs.length / 2;
+                    for (const index in logs) {
```
kkuzmin, line 453: it should be easier to use [`array.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) to split an array into halves.

kkuzmin: for example you can have a list of indexes to be handled in a `map`, like:
```
const indexArray = [
    {start: 0, stop: half}, 
    {start:half+1, logs.length}
];

indexArray.map(logsPart => {
    collector.batchLogProcess(logs.slice(logsPart.start, logsPart.stop, ....);
});

```

kkuzmin (changes requested):

paws_collector.js
```diff
@@ -447,6 +439,49 @@ class PawsCollector extends AlAwsCollector {
...
+                            halfLogs = [];
+                            loglength = logs.length;
+                            halfLogs.push(logs[index]);
+                            if (parseInt(index) == logs.length-1) {
+                                batchLogs.push(halfLogs);
+                            }
+                        }
+                    }
+                    batchLogs.map(batchLog => {
+                         collector.batchLogProcess(batchLog, privCollectorState, nextInvocationTimeout, asyncCallback);
```
kkuzmin, line 471: we can't call `asyncCallback` here, becasue once the first batch is sent the callback will be returned and invocation will end without waiting for the 2nd batch to be sent. 
We need to send batches one by one and probably use `array.reduce()`

paws_collector.js
kkuzmin: and use `indexArray.reduce()` instead of `indexArray.map()`

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: indexArray.reduce() return the single value.  I am checking  foe Async.reduce().

paws_collector.js
rakhimundhada15: Using async.reduce().

paws_collector.js
```diff
@@ -495,6 +487,37 @@ class PawsCollector extends AlAwsCollector {
...
+                 */
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
+
+                    let half = logs.length / 2;
+                    const indexArray = [
+                        { start: 0, stop: half },
+                        { start: half, stop: logs.length }
+                    ];
+
+                    async.reduce(indexArray, 0, (memo, logpart, asyncCallback) => {
```
kkuzmin, line 507: reduce returns a promise if 4th param `callback` is not pass
We need to add a new callback here and check the output of each `batchLogProcess`

kkuzmin: I think we can also use `map()` here, but I don't have strong preference.

kkuzmin (changes requested):

paws_collector.js
```diff
@@ -495,6 +487,37 @@ class PawsCollector extends AlAwsCollector {
...
+        return collector.processLog(logs, collector.pawsFormatLog.bind(collector), null,  (err) => {
+           
+            if (err) {
+                /**
+                 * MS api return the logs but if build payload size > 10MB, and split logs array the in two part.
+                 * Call the processLog again to send Logmsgs to ingest. 
+                 */
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
+
+                    let half = logs.length / 2;
```
kkuzmin, line 501: please make it `const`
and also wrap it into `Math.floor()`

kkuzmin (changes requested):

paws_collector.js
```diff
@@ -495,6 +487,37 @@ class PawsCollector extends AlAwsCollector {
...
+                    const indexArray = [
+                        { start: 0, stop: half },
+                        { start: half, stop: logs.length }
```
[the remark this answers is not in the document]
rakhimundhada15: array.slice exclude(stop: half)  last index value, so including that value in next half(start: half)

paws_collector.js
```diff
@@ -495,6 +487,48 @@ class PawsCollector extends AlAwsCollector {
...
+                                    reject(err);
+                                }
+                                else {
+                                    resolve(res);
+                                }
+                            });
+                        });
+                    });
+                    const results = await Promise.all(promises);
+                    return callback(null, results[results.length-1], nextInvocationTimeout);
```
kkuzmin, line 520: I think we need to return `privCollectorState ` instead of `results[results.length-1]`

rakhimundhada15: both are returning same value. I will use privCollectorState.

kkuzmin (approved):

PR #204, opened 2021-07-30
Catch JSON parse exceptions

kkuzmin:
### Problem Description
Exception is happening in asynchronous function hence no catch in `Promise.catch()`

### Solution Description
Catch parsing exception inside asynchronous function and reject promise.


OHaganA (approved):
Simple change, LGTM

PR #205, opened 2021-08-02
Process the logmsgs in batch if log messages size greater than 10K

rakhimundhada15:
### Problem Description
For few o365 collector we found ingest error, max payload issue and other issues if data is huge . So process logs in batches.

### Solution Description
Split the data per 10K message size and create batch. Call the processLog to send Logmsgs to ingest.

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -523,7 +523,7 @@ class PawsCollector extends AlAwsCollector {
                     });
                     await Promise.all(promises);
                     return callback(null, privCollectorState, nextInvocationTimeout);
-                } else if (err.statusCode && err.statusCode === 307 && process.env.paws_collection_interval && process.env.paws_collection_interval > 0) {
+                } else if (err.statusCode && err.statusCode === 307 && process.env.paws_collection_interval && process.env.paws_collection_interval > 1) {
                     const paws_collection_interval = Math.floor(process.env.paws_collection_interval / 2);
 
                     AlAwsUtil.setEnv({ paws_collection_interval: paws_collection_interval }, (err, res) => {
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -523,7 +523,14 @@ class PawsCollector extends AlAwsCollector {
                     });
                     await Promise.all(promises);
                     return callback(null, privCollectorState, nextInvocationTimeout);
-                } else {
+                } else if (err.statusCode && err.statusCode === 307 && process.env.paws_collection_interval && process.env.paws_collection_interval > 0) {
+                    const paws_collection_interval = Math.floor(process.env.paws_collection_interval / 2);
+
+                    AlAwsUtil.setEnv({ paws_collection_interval: paws_collection_interval }, (err, res) => {
+                        return callback(null, privCollectorState, nextInvocationTimeout);
+                    });
+                }
+                else {
                     collector.reportErrorToIngestApi(err, () => {
                         return callback(err);
                     });
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -503,14 +503,11 @@ class PawsCollector extends AlAwsCollector {
     batchLogProcess(logs, privCollectorState, nextInvocationTimeout, callback) {
         let collector = this;
         let indexArray = [];
-        if (logs.length > MAX_LOG_BATCH_SIZE) {
-            const batches = Math.ceil(logs.length / MAX_LOG_BATCH_SIZE);
-            for (let i = 0; i < batches; i++) {
-                indexArray.push({ start: MAX_LOG_BATCH_SIZE * i, stop: MAX_LOG_BATCH_SIZE * (i + 1) });
-            }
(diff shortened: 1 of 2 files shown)


paws_collector.js
```diff
@@ -523,7 +523,14 @@ class PawsCollector extends AlAwsCollector {
                     });
                     await Promise.all(promises);
                     return callback(null, privCollectorState, nextInvocationTimeout);
-                } else {
+                } else if (err.statusCode && err.statusCode === 307 && process.env.paws_collection_interval && process.env.paws_collection_interval > 0) {
+                    const paws_collection_interval = Math.floor(process.env.paws_collection_interval / 2);
```
scott-jones-AlertLogic, line 527: I think we should introduce some lower bound for `paws_collection_interval`.  If `process.env.paws_collection_interval` is 1 when it hits this clause, we would end up with a new collection interval of 0
```javascript
> process.env.paws_collection_interval = 1
1
> Math.floor(process.env.paws_collection_interval / 2)
0
```

paws_collector.js
```diff
@@ -498,17 +499,20 @@ class PawsCollector extends AlAwsCollector {
...
                  */
-                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1)) {
-
-                    const half = Math.floor(logs.length / 2);
-                    const indexArray = [
-                        { start: 0, stop: half },
-                        { start: half, stop: logs.length }
-                    ];
-
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1) || (err.statusCode && err.statusCode === 307)) {
```
kkuzmin, line 506: I suggest we do split regardless of this `if` condition. Basically move the code below to [here](https://github.com/alertlogic/paws-collector/pull/205/files#diff-82fd638370f7c86cddbb67ebf81bc2952ba84a1e7dc12ce1ed0fe1da30045492R497)

paws_collector.js
```diff
@@ -38,6 +38,7 @@ const DDB_OPTIONS = {
     ConsistentRead: true
 };
 
+const MAX_LOG_SIZE = 10000;
```
kkuzmin, line 41: please rename it to something like `MAX_LOG_BATCH_SIZE`

paws_collector.js
```diff
@@ -498,17 +499,20 @@ class PawsCollector extends AlAwsCollector {
...
-                        { start: 0, stop: half },
-                        { start: half, stop: logs.length }
-                    ];
-
+                if ((typeof err === 'string' && err.indexOf('Maximum payload size exceeded') !== -1) || (err.statusCode && err.statusCode === 307)) {
+                    const indexArray = [];
+                    if (logs.length > MAX_LOG_SIZE) {
+                        const oneBatch = Math.ceil(logs.length / MAX_LOG_SIZE);
+                        for (let i = 0; i < oneBatch; i++) {
+                            indexArray.push({ start: MAX_LOG_SIZE * i, stop: MAX_LOG_SIZE * (i + 1) });
```
kkuzmin, line 511: I suggest we use `array.slice()`, like [here](https://stackoverflow.com/questions/8495687/split-array-into-chunks)

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: Here i am just forming the array base on MAX_LOG_BATCH_SIZE.

paws_collector.js
```diff
@@ -491,46 +492,44 @@ class PawsCollector extends AlAwsCollector {
...
-                    return callback(null, privCollectorState, nextInvocationTimeout);
-                } else {
-                    collector.reportErrorToIngestApi(err, () => {
-                        return callback(err);
-                    });
-                }
-            } else {
-                return callback(null, privCollectorState, nextInvocationTimeout);
+        let indexArray = [];
+        if (logs.length > MAX_LOG_BATCH_SIZE) {
```
kkuzmin, line 506: this `if..else` statement is unnecessary. if `slice.stop` is more than array length then the whole array is returned.

kkuzmin (changes requested):

test/paws_test.js
```diff
@@ -598,8 +641,7 @@ describe('Unit Tests', function() {
                     done();
                 },
                 succeed: function () {
-                    AWS.restore('DynamoDB');
-                    alserviceStub.alog.restore();
+
```
kkuzmin, line 644: could you please check here that send logs function was called 3 times.

kkuzmin (changes requested):

test/paws_test.js
rakhimundhada15: sure

rakhimundhada15: Checked the sendLogmsgs and sendLmcstats function called thrice

kkuzmin (approved):

PR #210, opened 2021-08-18
 Reduce the fluctuations rate of collector status

rakhimundhada15:
### Problem Description
If there is intermediate  errors then collector status fluctuating from healthy to unhealthy and vice versa

### Solution Description
Added the retry_count in SQS state and increment the value if error occurred . If retry_count reached to 5 then only send the error status to assets to reduce the fluctuations of collector status.

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -463,7 +463,7 @@ class PawsCollector extends AlAwsCollector {
                             // Here we just upload same state messaged into SQS and  send error status to the backend.
                             // The invocation is marked as succeed in order to clean up current state message in SQS.
                             // Increment/reset the retry count if error occured. 
-                            pawsState.retry_count = (!pawsState.retry_count || pawsState.retry_count >= 5) ? 1 : pawsState.retry_count + 1;
+                            pawsState.retry_count = (!pawsState.retry_count || pawsState.retry_count >= MAX_ERROR_RETRIES) ? 1 : pawsState.retry_count + 1;
                             collector._storeCollectionState(pawsState, pawsState.priv_collector_state, 300, function(storeError){
                                 if (!storeError){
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -460,6 +462,8 @@ class PawsCollector extends AlAwsCollector {
                             // in order to avoid expiration of that message due to retention period.
                             // Here we just upload same state messaged into SQS and  send error status to the backend.
                             // The invocation is marked as succeed in order to clean up current state message in SQS.
+                            // Increment/reset the retry count if error occured. 
+                            pawsState.retry_count = (!pawsState.retry_count || pawsState.retry_count >= 5) ? 1 : pawsState.retry_count + 1;
```
kkuzmin, line 466: please use `MAX_ERROR_RETRIES `

kkuzmin (changes requested):
I'd be nice to have a unit test for it

test/paws_test.js
```diff
@@ -674,6 +674,85 @@ describe('Unit Tests', function() {
...
+                });
+
+            let mockPawsGetLogs = sinon.stub(TestCollector.prototype, "pawsGetLogs").callsFake(
+                function fakeFn(state, callback) {
+                    return callback({ name: 'OktaApiError', status: 401, errorCode: 'E0000011', errorSummary: 'Invalid token provided' });
+                });
+            const testEvent = {
+                Records: [
+                    {
+                        "body": "{\n  \"priv_collector_state\": {\n    \"since\": \"2021-07-01T02:37:37.617Z\",\n    \"until\": \"2021-07-01T03:37:37.617Z\"\n  }\n, \"retry_count\":3}",
```
kkuzmin, line 745: please add one more test case where state doesn't have `retry` property. To cover existing collectors.

kkuzmin (changes requested):

kkuzmin (approved):

PR #212, opened 2021-09-02
Changed polling interval from n-cap to hour-cap as per suggestion

sameerbag:
Changed polling interval from n-cap to hour-cap as per suggestion


rakhimundhada15 (approved):

rakhimundhada15:
Bump the package version before merge.

PR #213, opened 2021-09-07
Added the custom health check for Paws collector

rakhimundhada15:
### Problem Description
Checkin function update the status to ok if there is no error on Lambda Error metrics.
Paws collector not sending error to Lambda error metrics in few scenario ;so Checkin function update the status to 'ok' even we have error for collector. But paws collector send the error to custom metrics like [PawsClientError](https://github.com/alertlogic/paws-collector/blob/master/paws_collector.js#L414)/[PawsIngestError](https://github.com/alertlogic/paws-collector/blob/master/paws_collector.js#L537) .

### Solution Description
Fetch the data from `PawsClientError` metrics using getCustomMetrics and send as health check to [al-aws-collector.js](https://github.com/alertlogic/al-aws-collector-js/blob/master/al_aws_collector.js#L128) which will check the status in [ checkin function ](https://github.com/alertlogic/al-aws-collector-js/blob/master/al_aws_collector.js#L423).
If lambda error metrics and PawsClientError metrics both did not show any error then only we will send the status ok to assets.

### DIFF
--- a/paws_health_checks.js
+++ b/paws_health_checks.js
@@ -28,7 +28,7 @@ function customHealthCheck(asyncCallback) {
         process.env.AWS_LAMBDA_FUNCTION_NAME, 'PawsClientError', 'PawsCollectors', customDimentions, (err, res) => {
             const datapointSum = res && res.Datapoints.length > 0 && res.Datapoints[0].Sum ? res.Datapoints[0].Sum : 0;
             if (datapointSum > 0) {
-                return asyncCallback(al_health.errorMsg('PAWS000403', 'Paws client errors'));
+                return asyncCallback(al_health.errorMsg('PAWS000403','Something wrong with streams or client configuration'));
             }
             return asyncCallback(null);
         }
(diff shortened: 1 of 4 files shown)


paws_health_checks.js
```diff
@@ -0,0 +1,40 @@
...
+    {
+        Name: 'CollectorType',
+        Value: process.env.paws_type_name
+    }
+
+    AlAwsStatsTmpls.getCustomMetrics(
+        process.env.AWS_LAMBDA_FUNCTION_NAME, 'PawsClientError', 'PawsCollectors', customDimentions, (err, res) => {
+            const datapointSum = res && res.Datapoints.length > 0 && res.Datapoints[0].Sum ? res.Datapoints[0].Sum : 0;
+            if (datapointSum > 0) {
+                return asyncCallback(al_health.errorMsg('PAWS000403','Something wrong with streams or client configuration'));
```
kkuzmin, line 31: Please say `Please check stream status errors and collector configuration`

kkuzmin (approved):

PR #216, opened 2021-09-20
Added Poll interval delay for paws collector

rakhimundhada15:
### Problem Description

Few  Paws collector missed the data.This issue we seen for Cisco duo collector.
Scenario : When we made the api call just within the 1 min( at 7.43.59 pm we try to fetch the data for duration of 7.41.27pm to 7.42.27pm.)  
There may be chance of complete data was not available at that time when we make api call within a min.

### Solution Description
Make the api call for the time stamp of 10 to 15 min ago.

Set the NEXT_POLL_INTERVAL_DELAY to 10 to 15 min If current time and nextUntilMoment difference is less the NEXT_POLL_INTERVAL_DELAY seconds.

### DIFF
--- a/paws_utils.js
+++ b/paws_utils.js
@@ -12,7 +12,7 @@
 const moment = require('moment');
 
 function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
-    const NEXT_POLL_INTERVAL_DELAY = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 300;
+    const pollIntervalDelay = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 300;
     const nowMoment = moment();
     const nextSinceMoment = curUntilMoment.isAfter(nowMoment) ?
         nowMoment : curUntilMoment;
@@ -71,14 +71,14 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
      * make next API call after 5 to 10 min to avoid any loss of data.
      */
     let nextPollInterval;
-    if (nowMoment.diff(nextUntilMoment, 'hours') > 1) {
+    if (nowMoment.diff(nextUntilMoment, 'minutes') > 30) {
         nextPollInterval = 1;
     }
-    else if (nowMoment.diff(nextUntilMoment, 'seconds') > NEXT_POLL_INTERVAL_DELAY) {
+    else if (nowMoment.diff(nextUntilMoment, 'seconds') > pollIntervalDelay) {
         nextPollInterval = pollInterval;
     }
     else {
-        nextPollInterval = NEXT_POLL_INTERVAL_DELAY;
+        nextPollInterval = pollIntervalDelay;
     }
 
     return { nextSinceMoment, nextUntilMoment, nextPollInterval };
--- a/paws_utils.js
+++ b/paws_utils.js
@@ -10,6 +10,7 @@
 'use strict';
 
 const moment = require('moment');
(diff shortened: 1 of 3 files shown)


paws_utils.js
```diff
@@ -64,9 +65,17 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
...
     }
 
-    const nextPollInterval = nowMoment.diff(nextUntilMoment, 'seconds') > pollInterval ?
-            1 : pollInterval;
+    /**
+     * If current time and nextUntilMoment difference is less the NEXT_POLL_INTERVAL_DELAY seconds,
+     * then pull the data for 1 min to keep the poll interval delay consistance and get whole data and not missed anything.
+     */
+    if (nowMoment.diff(nextUntilMoment, 'seconds') < NEXT_POLL_INTERVAL_DELAY) {
+        nextUntilMoment = moment(nextSinceMoment).add(60, 'seconds');
```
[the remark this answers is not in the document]
kkuzmin: And also check that `nextSinceMoment` is still less than `nextUntilMoment` after modification. If it happens that `nextSinceMoment > nextUntilMoment ` then lets wait and have `nextSinceMoment = nextUntilMoment`

paws_utils.js
```diff
@@ -64,9 +65,17 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
...
+    }
 
+    // If current time and nextUntilMoment difference is less the 10 min then set nextPollInterval to 10 min(600 sec) else 1 sec.
+    const nextPollInterval = nowMoment.diff(nextUntilMoment, 'seconds') > NEXT_POLL_INTERVAL_DELAY ?
```
kkuzmin, line 77: Please refactor `nowMoment.diff()` calculation into a separate var. I see it is used in multiple places.

paws_utils.js
```diff
@@ -10,6 +10,7 @@
 'use strict';
 
 const moment = require('moment');
+const NEXT_POLL_INTERVAL_DELAY = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 600;
```
kkuzmin, line 13: I think we need to move this var inside `calcNextCollectionInterval()` and default its value to `pollInterval` and not `600`

kkuzmin (changes requested):

paws_utils.js
rakhimundhada15: The default  value of pollInterval in all the collector is 60seconds. If we used this it will just keep 1 min interval ,it will be the same as existing logic.

paws_utils.js
```diff
@@ -12,45 +12,47 @@
 const moment = require('moment');
 
 function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
+    const NEXT_POLL_INTERVAL_DELAY = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 300;
```
kkuzmin, line 15: please fix variable name.

paws_utils.js
```diff
@@ -64,8 +66,20 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
...
     }
 
-    const nextPollInterval = nowMoment.diff(nextUntilMoment, 'seconds') > pollInterval ?
-            1 : pollInterval;
+    /**
+     * If current moment and nextUntilMoment difference is less than given poll_interval_delay then set nextPollInterval = poll_interval_delay;
+     * make next API call after 5 to 10 min to avoid any loss of data.
+     */
+    let nextPollInterval;
+    if (nowMoment.diff(nextUntilMoment, 'hours') > 1) {
```
kkuzmin, line 74: let's have 30 mins here.

kkuzmin (changes requested):
+1 in general. looks a lot simpler. But some minor comments.

paws_utils.js
rakhimundhada15: changed to pollIntervalDelay.

kkuzmin (approved):

PR #217, opened 2021-09-27
Poll interval delay should not increased more than 15 min

rakhimundhada15:
### Problem Description
Currently  we keep minimum 10 min delay and it increase up to 30 min as we keep 60sec interval after each call.

### Solution Description
If interval increase more than 15 min then we make set 1secs interval for each call.

### DIFF
--- a/paws_utils.js
+++ b/paws_utils.js
@@ -12,6 +12,7 @@
 const moment = require('moment');
 
 function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
+    const INTERVAL_IN_MINUTES = 15;
     const pollIntervalDelay = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 300;
     const nowMoment = moment();
     const nextSinceMoment = curUntilMoment.isAfter(nowMoment) ?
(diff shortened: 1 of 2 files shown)


paws_utils.js
```diff
@@ -71,7 +71,7 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
      * make next API call after 5 to 10 min to avoid any loss of data.
      */
     let nextPollInterval;
-    if (nowMoment.diff(nextUntilMoment, 'minutes') > 30) {
+    if (nowMoment.diff(nextUntilMoment, 'minutes') > 15) {
```
kkuzmin, line 74: please use a const variable.

kkuzmin (changes requested):
Please bump package version

paws_utils.js
```diff
@@ -12,6 +12,7 @@
 const moment = require('moment');
 
 function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
+    const INTERVAL_IN_MINUTES = 15;
```
kkuzmin, line 15: I think we need to call it `DELAY_IN_MINUTES`

kkuzmin (changes requested):

kkuzmin (approved):

PR #219, opened 2021-09-30
Use unique message ID as dedup table key

kkuzmin:
### Problem Description
Deduplication DDB table doesn't help when collector state doesn't change because MD5 hash of a message is used as DDB item key.
 
### Solution Description
Use unique message id as dedup table key instead of md5 hash.

### DIFF
--- a/cfn/paws-collector.template
+++ b/cfn/paws-collector.template
@@ -279,28 +279,6 @@
                 ]
             }
         },
-        "BasicLambdaRole":{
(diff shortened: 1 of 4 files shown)


sameerbag (approved):

rakhimundhada15 (approved):
LGTM

rakhimundhada15 (approved):
BasicLambdaRole change looks good to me.

PR #220, opened 2021-10-01
AuthO : In case if nextLogId is null then set the state on basic of last collected Ts 

rakhimundhada15:
### Problem Description
IF last log id is null then it start from current moment instate of last collected time stamp.

### Solution Description
In initial call if we get nextLogId  null then set last collected time stamp and if last collect time is undefined then start with current time.

### DIFF
--- a/collectors/auth0/auth0_collector.js
+++ b/collectors/auth0/auth0_collector.js
@@ -72,9 +72,10 @@ class Auth0Collector extends PawsCollector {
                 1 : this.pollInterval;
 
         if (nextLogId === null) {
-            //If collector initial call and get empty logs then this condition will work
+            // If collector initial call and get empty logs then this condition will work
+            const lastTs = lastLogTs ? lastLogTs : curState.since;
             return {
-                since: curState.since,
+                since: lastTs,
                 poll_interval_sec: 1
             };
         }
--- a/collectors/auth0/utils.js
+++ b/collectors/auth0/utils.js
@@ -2,8 +2,8 @@ const moment = require('moment');
 
 function getAPILogs(auth0Client, state, accumulator, maxPagesPerInvocation) {
     let pageCount = 0;
-
-    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date:[" + moment(state.since).toISOString() + " TO *]", per_page:100, sort: "date:1" };
+    const lastTs = state.last_collected_ts ? state.last_collected_ts : state.since;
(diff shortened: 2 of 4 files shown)


collectors/auth0/auth0_collector.js
```diff
@@ -71,10 +71,11 @@ class Auth0Collector extends PawsCollector {
         const nextPollInterval = nowMoment.diff(lastLogMoment, 'seconds') > this.pollInterval ?
                 1 : this.pollInterval;
 
-        if (nextLogId === null) {
-            //If collector initial call and get empty logs then this condition will work
-            return {
-                since: curState.since,
+        if(nextLogId === null){
+              //If collector initial call and get empty logs then this condition will work
+             let lastTs = lastLogTs ? lastLogTs : curState.since;
```
kkuzmin, line 76: please use `const`

collectors/auth0/utils.js
```diff
@@ -2,8 +2,8 @@ const moment = require('moment');
 
 function getAPILogs(auth0Client, state, accumulator, maxPagesPerInvocation) {
     let pageCount = 0;
-
-    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date:[" + moment(state.since).toISOString() + " TO *]", per_page:100, sort: "date:1" };
+    let lastTs = state.last_collected_ts ? state.last_collected_ts : state.since;
+    let params = state.last_log_id ? { from: state.last_log_id, take:100 } : { q: "date:[" + moment(lastTs).toISOString() + " TO *]", per_page:100, sort: "date:1" };
```
kkuzmin, line 6: please use `const`

kkuzmin (changes requested):

collectors/auth0/auth0_collector.js
```diff
@@ -71,10 +71,11 @@ class Auth0Collector extends PawsCollector {
         const nextPollInterval = nowMoment.diff(lastLogMoment, 'seconds') > this.pollInterval ?
                 1 : this.pollInterval;
 
-        if (nextLogId === null) {
-            //If collector initial call and get empty logs then this condition will work
-            return {
-                since: curState.since,
+        if(nextLogId === null){
+              //If collector initial call and get empty logs then this condition will work
+              const lastTs = lastLogTs ? lastLogTs : curState.since;
```
kkuzmin, line 76: please fix indention

kkuzmin (approved):

PR #224, opened 2021-10-14
Cisco Duo : Fix unix timestamp format which display in logs

rakhimundhada15:
### Problem Description

1. Unix timestamp not formatted correctly while display for Cisco duo in cloud  watch log 
ex. `Administrator from 1970-01-19T20:55:07.533Z` 
2. If there is no logs for given time stamp we set same timestamp as next page and  try to poll the data for same again and again until we get the data.

### Solution Description
1.Converted unix timestamp to timestamp in milliseconds and then show in ISO format .
2.As per cisco [duo api document](https://duo.com/docs/adminapi#offline-enrollment-logs),For Offline Enrollment stream, Return the list if **events ranging from the last 180 days up to as recently as two minute**s before the API request. So considering  similar behaviour  for other two stream[Telephony Logs, Administrator Logs] . 
If we don't get the data for last provided mintime stamp we will move the window to (now-1hr) and if (now-1hr) < mintime then move window by 1 secs.

### DIFF
--- a/collectors/ciscoduo/utils.js
+++ b/collectors/ciscoduo/utils.js
@@ -42,7 +42,7 @@ function getAPILogs(client, objectDetails, state, accumulator, maxPagesPerInvoca
                                 // If there is no data check if min time stamp is less than 1 hr move by 1 sec else set the last hour time stamp as nextPage
                                 const lastHourMoment = moment().subtract(1, 'hours').unix();
                                 const mintime = parseInt(objectDetails.query.mintime);
-                                nextPage = lastHourMoment > mintime ? lastHourMoment : mintime + 1;
+                                nextPage = Math.max(mintime + 1, lastHourMoment);
                                 return resolve({ accumulator, nextPage });
                             }
                         }


collectors/ciscoduo/utils.js
```diff
@@ -39,7 +39,8 @@ function getAPILogs(client, objectDetails, state, accumulator, maxPagesPerInvoca
                                 getData();
                             }
                             else {
-                                nextPage = objectDetails.query.mintime;
+                                // Moving time stamp by 1 sec if there is no records;
+                                nextPage = parseInt(objectDetails.query.mintime) + 1;
```
kkuzmin, line 43: I don't quite get why we move it by 1 sec. What is the time interval we poll here for? Should we move move to the end of previous poll interval minus 10 mins?

kkuzmin (changes requested):

collectors/ciscoduo/utils.js
rakhimundhada15: Adding 60 sec to move the window

collectors/ciscoduo/collector.js
```diff
@@ -161,10 +162,14 @@ class CiscoduoCollector extends PawsCollector {
             };
         } else {
             //There is no next page concept for this API, So Setting up the next state mintime using the last log (Unix timestamp + 1).
+            const nowMoment = moment();
+            const nextMintime = moment(parseInt(nextPage * 1000));
```
kkuzmin, line 166: So do we have timestamp in seconds or milliseconds?

collectors/ciscoduo/utils.js
```diff
@@ -39,7 +39,8 @@ function getAPILogs(client, objectDetails, state, accumulator, maxPagesPerInvoca
                                 getData();
                             }
                             else {
-                                nextPage = objectDetails.query.mintime;
+                                // Moving time stamp by 60 sec if there is no records;
+                                nextPage = parseInt(objectDetails.query.mintime) + 60;
```
kkuzmin, line 43: if `mintime` is in milliseconds then it is going to be slow.

collectors/ciscoduo/collector.js
```diff
@@ -98,8 +98,9 @@ class CiscoduoCollector extends PawsCollector {
         tsPaths = objectDetails.tsPaths;
 
         const stateMaxtime = state.maxtime ? `till ${moment(parseInt(state.maxtime)).toISOString()}` : ``;
+        const stateMintime = state.stream === Authentication ? moment(parseInt(state.mintime)).toISOString() : moment(parseInt(state.mintime * 1000)).toISOString(); // Convert Epoch timestamp to milliseconds to get date in correct format
```
kkuzmin, line 101: should we do the same for `stateMaxtime ` above?

collectors/ciscoduo/collector.js
```diff
@@ -161,10 +162,15 @@ class CiscoduoCollector extends PawsCollector {
             };
         } else {
             //There is no next page concept for this API, So Setting up the next state mintime using the last log (Unix timestamp + 1).
+            // Added 5 min interval if there is no data and mintime is less than 5min. Wait for 5 min and then try again for next page.
+            const nowMoment = moment();
+            const nextMintime = moment(parseInt(nextPage * 1000));
+            const nextPollInterval = nowMoment.diff(nextMintime, 'seconds') > 300 ?
```
kkuzmin, line 168: I think we don't need it here because more data is available for collection. Let's keep using 1 sec here unless you've spotted any API throttling errors

kkuzmin (changes requested):

collectors/ciscoduo/collector.js
rakhimundhada15: maxtime we get only for Authentication stream. The value is already in milliseconds.

kkuzmin: should we have same conversion for `state.maxtime` above?

collectors/ciscoduo/utils.js
```diff
@@ -39,7 +39,10 @@ function getAPILogs(client, objectDetails, state, accumulator, maxPagesPerInvoca
                                 getData();
                             }
                             else {
-                                nextPage = objectDetails.query.mintime;
+                                // If there is no data check if min time stamp is less than 1 hr move by 1 sec else set the last hour time stamp as nextPage
+                                const lastHourMoment = moment().subtract(1, 'hours').unix();
+                                const mintime = parseInt(objectDetails.query.mintime);
+                                nextPage = lastHourMoment > mintime ? lastHourMoment : mintime + 1;
```
kkuzmin, line 45: let's use `Math.min(mintime+1, lastHourMoment)` here

kkuzmin (changes requested):

collectors/ciscoduo/utils.js
rakhimundhada15: Used Math.max(mintime+1, lastHourMoment) to get  latest time.

kkuzmin (approved):

collectors/ciscoduo/collector.js
rakhimundhada15: Maxtime we get only for Authentication stream which we format at line100 and for rest streams[Administrator, Telephony, OfflineEnrollment] we set maxtime empty `''` to display.

PR #234, opened 2021-11-15
change to handle no error code

sameerbag:
### Problem Description
No error code was not handle

### Solution Description
Added an error code using which datadog can represent that error code in the graph.

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -85,7 +85,7 @@ class OktaCollector extends PawsCollector {
     }
 
     _isNoErrorCode(error) {
-        return error.errorCode ? error.errorCode : "ale00011";
+        return error.errorCode ? error.errorCode : error.status;
     }
 
     _isThrottlingError(error) {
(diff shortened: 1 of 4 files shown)


collectors/okta/okta_collector.js
```diff
@@ -83,6 +84,10 @@ class OktaCollector extends PawsCollector {
         });
     }
 
+    _isNoErrorCode(error) {
+        return error.errorCode ? error.errorCode : "ale00011";
```
rakhimundhada15, line 88: Could you please check do we get errorCode field in error object or it is error.status as mentioned  at 
 Throttling https://github.com/alertlogic/paws-collector/pull/234/files#diff-7857e3061ff43ce02f1e5856dda622d45520533da0adbdc3509b3d98d4b74031R92

rakhimundhada15:
Also add the Unit test case of this.

collectors/okta/okta_collector.js
sameerbag: Added status code instead of errorcode

sameerbag: Added test case for the same

kkuzmin (changes requested):
Please bump package version

sameerbag:
@kkuzmin : Done, bumped package.json version

kkuzmin (approved):

PR #235, opened 2021-11-18
[o365]: Reduce the multiple SSM call to get credentials 

rakhimundhada15:
### Problem Description
Made the multiple call to SSM to get credentials and some time it will take time/face issue to return the credentials .
So we get error `Error: secret must be a non empty string` .

### Solution Description

1. Remove the PawsCollector.load() call from getO365ManagmentClient() to reduce the multiple call to SSM param.
2. Save the paws credentials in one object  when constructor get call and pass to O365 api from different  method (getPreFormedUrl, subscriptionsContent).

### DIFF
--- a/collectors/o365/healthcheck.js
+++ b/collectors/o365/healthcheck.js
@@ -12,13 +12,14 @@
 
 const o365_mgmnt = require('./lib/o365_mgmnt');
 const al_health = require('@alertlogic/al-aws-collector-js').Health
-
+const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
 /*
  * Checks the subscriptions against the configured content type. Starts the subscriptions if th
  */
 
 function checkO365Subscriptions(callback){
-    return o365_mgmnt.listSubscriptions()
+    return PawsCollector.load().then(({pawsCreds}) => {
+        return o365_mgmnt.listSubscriptions(pawsCreds)
         .then(filterSubscriptions)
         .then(filteredStreams => {
             if(filteredStreams.length > 0){
@@ -26,7 +27,7 @@ function checkO365Subscriptions(callback){
             } else{
                 console.info(`O365000102: No streams need restarted.`);
             }
-            const streamPromises = filteredStreams.map(stream => o365_mgmnt.startSubscription(stream));
+            const streamPromises = filteredStreams.map(stream => o365_mgmnt.startSubscription(stream, pawsCreds));
             return Promise.all(streamPromises);
         })
         .then(res => callback(null))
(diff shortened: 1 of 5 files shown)


collectors/o365/lib/o365_mgmnt/index.js
```diff
@@ -57,8 +57,8 @@ var _listSubscriptions = function() {
  * @returns {Promise}
  *
  */
-var _startSubscription = function(contentType) {
-    return getO365ManagmentClient().then((client) => {
+var _startSubscription = function(contentType, pawsCreds) {
```
kkuzmin, line 60: can we please have `pawsCreds` as the first argument in every function

collectors/o365/healthcheck.js
```diff
@@ -12,21 +12,22 @@
...
 const al_health = require('@alertlogic/al-aws-collector-js').Health
-
+const PawsCollector = require('@alertlogic/paws-collector').PawsCollector;
 /*
  * Checks the subscriptions against the configured content type. Starts the subscriptions if th
  */
 
 function checkO365Subscriptions(callback){
-    return o365_mgmnt.listSubscriptions()
+    return PawsCollector.load().then(({pawsCreds}) => {
```
kkuzmin, line 21: according to this [code](https://github.com/alertlogic/al-azure-collector-js/blob/master/master.js#L309) first arg is collector wher eyou can probably get `pawsCreds` from.
I think we need to just bind health check function context so that we can use `this` inside it

kkuzmin (changes requested):

collectors/o365/healthcheck.js
kkuzmin: Context binding already done [here](https://github.com/alertlogic/al-aws-collector-js/blob/a22017e9e6033d16f8fcf7596039f79afc0d010a/al_aws_collector.js#L468). So you should be able to use `this.pawsCreds` inside this function

rakhimundhada15: sure. I will check and update.

collectors/o365/lib/o365_mgmnt/index.js
kkuzmin: actually, please rename `pawsCreds` to just `creds` or `azureCreds`

rakhimundhada15: used creds as first parameter

kkuzmin (approved):
@rakhimundhada15 thank you. This change solves the issue with credential load.
Lat's have a followup change to pass O365 management client as a first parameter to every o365 management function. Currently `getO365ManagmentClient()` is called every time we talk to O365 API which is odd. I think we can call it once an store inside collector object.

rakhimundhada15:
Sure , I will do the changes in separate pr.

PR #237, opened 2021-11-25
[Paws lib]: Bump the paws version to fix Lambda.updateFunctionConfiguration failures

rakhimundhada15:
### Problem Description
All paws deployment failed as updateFunctionConfiguration calls may fail if Lambda status is InProgress.

### Solution Description
Fix in al-aws-collector-js. [pr](https://github.com/alertlogic/al-aws-collector-js/pull/72 )

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alertlogic/paws-collector",
-  "version": "2.0.21",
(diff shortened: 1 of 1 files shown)


rakhimundhada15:
test it

kkuzmin (approved):

PR #239, opened 2021-12-10
Paws-Lib Correct the file encoding format if ssm-direct is true

rakhimundhada15:
### Problem Description
For 0365 we getting the error `secret must be a non empty string.`.
After read the credentials from SSM we stored it in file in base64 format. 

Here is the Scenario when base64 value get set to secret instead of actual string :  When credentials file is exist and PAWS_DECRYPTED_CREDS  is empty ,it will check for ssm-direct and if it  true then we directly set that value as [secret](https://github.com/alertlogic/paws-collector/blob/master/paws_collector.js#L72) which is base64 format and not a  actual string.

### Solution Description
Base on ssm-direct value(if true) stored the credentials in utf-8 format else stored in base64.

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -42,8 +42,9 @@ const MAX_ERROR_RETRIES = 5;
 const MAX_LOG_BATCH_SIZE = 10000;
 function getPawsParamStoreParam(){
     return new Promise((resolve, reject) => {
-        if(fs.existsSync(CREDS_FILE_PATH)){
-            return resolve(fs.readFileSync(CREDS_FILE_PATH));
+        if (fs.existsSync(CREDS_FILE_PATH) && fs.readFileSync(CREDS_FILE_PATH).length !== 0) {
+            let fileContent = process.env.ssm_direct ? fs.readFileSync(CREDS_FILE_PATH, 'utf-8') : fs.readFileSync(CREDS_FILE_PATH);
+            return resolve(fileContent);
         }
         var ssm = new AWS.SSM();
         var params = {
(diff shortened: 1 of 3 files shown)


rakhimundhada15:
test it

paws_collector.js
```diff
@@ -42,8 +42,9 @@ const MAX_ERROR_RETRIES = 5;
 const MAX_LOG_BATCH_SIZE = 10000;
 function getPawsParamStoreParam(){
     return new Promise((resolve, reject) => {
-        if(fs.existsSync(CREDS_FILE_PATH)){
-            return resolve(fs.readFileSync(CREDS_FILE_PATH));
+        if (fs.existsSync(CREDS_FILE_PATH) && fs.readFileSync(CREDS_FILE_PATH).length !== 0) {
+            let fileContent = process.env.ssm_direct ? fs.readFileSync(CREDS_FILE_PATH, 'utf-8') : fs.readFileSync(CREDS_FILE_PATH);
```
kkuzmin, line 46: should we `fs.readFileSync(CREDS_FILE_PATH, 'base64')` if `ssm_direct  == false`?

kkuzmin (changes requested):

paws_collector.js
```diff
@@ -42,8 +42,9 @@ const MAX_ERROR_RETRIES = 5;
 const MAX_LOG_BATCH_SIZE = 10000;
 function getPawsParamStoreParam(){
     return new Promise((resolve, reject) => {
-        if(fs.existsSync(CREDS_FILE_PATH)){
-            return resolve(fs.readFileSync(CREDS_FILE_PATH));
+        if (fs.existsSync(CREDS_FILE_PATH) && fs.readFileSync(CREDS_FILE_PATH).length !== 0) {
```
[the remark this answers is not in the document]
rakhimundhada15: I will check and update.

paws_collector.js
rakhimundhada15: I will update this too. I have check we are getting base64, but for be specific i will add base64.

kkuzmin (approved):

PR #240, opened 2021-12-13
stored the ssm value in right encoding format

rakhimundhada15:
### Problem Description
What was fixed? - root cause

Read buffer string and pass to kms decrypt method to get actual  credentials  value.


rakhimundhada15:
closing it

PR #244, opened 2021-12-16
O365 : Create the o365 MS object once and used while making all api call

rakhimundhada15:
Instant of creating new MS management object every time we make each api call , create the object once and use same for all api call.

### DIFF
--- a/collectors/o365/lib/o365_mgmnt/index.js
+++ b/collectors/o365/lib/o365_mgmnt/index.js
@@ -15,19 +15,14 @@ const m_o365mgmnt = require('./o365management');
 const {ApplicationTokenCredentials} = require('@azure/ms-rest-nodeauth');
 
 function getO365ManagmentClient(creds) {
-    return new Promise((resolve, reject) => {
-        const AzureTenantId = process.env.paws_collector_param_string_1;
-        var g_appAdCreds = new ApplicationTokenCredentials(
-            creds.client_id,
-            AzureTenantId,
-            creds.secret,
-            'https://manage.office.com'
-        );
-        if (creds.client_id && creds.secret && AzureTenantId) {
-            resolve(new m_o365mgmnt.O365Management(g_appAdCreds, AzureTenantId));
-        }
-        else reject("Paws credential or AzureTenantId should not be empty")
-    });
+    const AzureTenantId = process.env.paws_collector_param_string_1;
+    var g_appAdCreds = new ApplicationTokenCredentials(
+        creds.client_id,
+        AzureTenantId,
+        creds.secret,
+        'https://manage.office.com'
+    );
+    return new m_o365mgmnt.O365Management(g_appAdCreds, AzureTenantId);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


collectors/o365/lib/o365_mgmnt/index.js
```diff
@@ -36,13 +34,12 @@ function getO365ManagmentClient(creds) {
  * Office 365 Management API subscription/content.
  * {@link https://docs.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-current-subscriptions Reference.}
  * 
+ * @param {O365Management} client - MSAzureObject for communicating with O365 management APIs
  * @returns {Promise}
  *
  */
-var _listSubscriptions = function(creds) {
```
kkuzmin, line 42: I think we can remove all the function down below bcause consumets can call client function directly, like `client.listSubscriptions(null)`.

collectors/o365/o365_collector.js
```diff
@@ -129,7 +125,7 @@ class O365Collector extends PawsCollector {
 
             if(nextPageUri && pageCount < process.env.paws_max_pages_per_invocation){
                 pageCount++;
-                return m_o365mgmnt.getPreFormedUrl(collector.pawsCred, nextPageUri)
+                return m_o365mgmnt.getPreFormedUrl(collector.o365_mgmnt_client, nextPageUri)
```
kkuzmin, line 128: I think we can call it like `collector.o365_mgmnt_client. getPreFormedUrl(nextPageUri)`

kkuzmin (changes requested):
@rakhimundhada15 thanks. I think we can do even more clean up and simplification

collectors/o365/lib/o365_mgmnt/index.js
```diff
@@ -8,102 +8,24 @@
...
             creds.secret,
             'https://manage.office.com'
         );
-        if (creds.client_id && creds.secret && AzureTenantId) {
-            resolve(new m_o365mgmnt.O365Management(g_appAdCreds, AzureTenantId));
-        }
-        else reject("Paws credential or AzureTenantId should not be empty")
-    });
+        return new m_o365mgmnt.O365Management(g_appAdCreds, AzureTenantId);
+    } else throw new Error("client secret or client id must be a non empty string.");
```
kkuzmin, line 26: do we have a top level try..catch statement in order to handle this exception?

kkuzmin (changes requested):

collectors/o365/index.js
```diff
@@ -15,7 +15,15 @@ const O365Collector = require('./o365_collector').O365Collector;
 exports.handler = O365Collector.makeHandler(function(event, context) {
     debug('input event: ', event);
     O365Collector.load().then(function(creds) {
-        var o365c = new O365Collector(context, creds);
-        o365c.handleEvent(event);
+        try {
+            var o365c = new O365Collector(context, creds);
+            o365c.handleEvent(event);
+        }
+        catch (err) {
```
kkuzmin, line 22: why do we need to catch and throw it again here?

rakhimundhada15: As per my understanding if we not add try catch then it will run next line too and then throw the error.
I will test and update.

kkuzmin (approved):

PR #247, opened 2022-01-07
[Auth0 and Cisco Duo Collectors] : Handle the Throttling/ Rate limit error

rakhimundhada15:
### Problem Description
Throttling error for Cisco duo and Auth0.
Error sample from Cisco duo collectors :
`PAWS000304 Error handling poll request: {\"code\":42901,\"message\":\"Too Many Requests\",\"stat\":\"FAIL\",\"errorCode\":42901}`

### Solution Description
 If we run into a rate limit/throttling error, instead of returning the error, we return the state back to the queue with an additional second.

1. Auth0 : Reference documents  for [ Auth0 Management API Endpoint Rate Limits](https://auth0.com/docs/support/policies/rate-limit-policy/management-api-endpoint-rate-limits). As per doc 10 api call per sec , but still added the 10 sec interval for safer side.
2. CiscoDuo : There is no proper document ,so refer the [ Cisco community doc](https://community.duo.com/t/429-too-many-requests-after-a-request-every-20-seconds/6863)  and added 60 sec poll interval as per comments mentioned on this.

### DIFF
--- a/collectors/ciscoduo/collector.js
+++ b/collectors/ciscoduo/collector.js
@@ -120,7 +120,7 @@ class CiscoduoCollector extends PawsCollector {
                 if (error.code && error.code === API_THROTTLING_ERROR) {
                     state.poll_interval_sec = state.poll_interval_sec < 60 ?
                         60 : state.poll_interval_sec + 1;
-                    AlLogger.warn(`The account has made too many requests of this type recently. Try again after ${state.poll_interval_sec} sec `)
+                    AlLogger.warn(`CDUO000003 API Request Limit Exceeded`, error);
                     collector.reportApiThrottling(function () {
                         return callback(null, [], state, state.poll_interval_sec);
                     });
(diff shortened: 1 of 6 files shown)


scott-jones-AlertLogic (approved):

OHaganA (approved):
Looks great Rahki, thanks

collectors/ciscoduo/collector.js
```diff
@@ -113,10 +113,23 @@ class CiscoduoCollector extends PawsCollector {
...
-                error.errorCode = error.code;
-                return callback(error);
+            }).catch((error) => {
+                // Cisco duo api has some rate limits that we might run into.
+                // If we run into a rate limit error, instead of returning the error,
+                // we return the state back to the queue with an additional  60 second added.
+                if (error.code && error.code === API_THROTTLING_ERROR) {
+                    state.poll_interval_sec = state.poll_interval_sec < 60 ?
+                        60 : state.poll_interval_sec + 1;
+                    AlLogger.warn(`The account has made too many requests of this type recently. Try again after ${state.poll_interval_sec} sec `)
```
kkuzmin, line 123: Let's report same warn log in Auth0 collector as well.

kkuzmin (approved):

PR #249, opened 2022-01-19
[Googlestackdriver Collector]Handle the RESOURCE_EXHAUSTED  error with message size greater than expected

rakhimundhada15:
### Problem Description
Throttling error with message size greater than expected. Below is the error message :
`Error: 8 RESOURCE_EXHAUSTED: Received message larger than max (4518352 vs. 4194304)`

### Solution Description:
 In case of error, half the time stamp interval and try again until the 15sec. If error still exist then skip or move the interval window by 15 sec.

### DIFF
--- a/collectors/googlestackdriver/collector.js
+++ b/collectors/googlestackdriver/collector.js
@@ -121,10 +121,21 @@ timestamp < "${state.until}"`;
                 // If we run inot a rate limit error, instead of returning the error,
                 // we return the state back to the queue with an additional second added, up to 10
                 // https://cloud.google.com/logging/quotas
+                // Error: 8 RESOURCE_EXHAUSTED: Received message larger than max (4518352 vs. 4194304),
+                // so half the given interval and if interval is less than 15 sec then move interval by same.
                 if(err.code === API_THROTTLING_ERROR){
                     const nextPollInterval = state.poll_interval_sec < 10 ?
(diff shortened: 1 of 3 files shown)


collectors/googlestackdriver/collector.js
```diff
@@ -121,10 +121,21 @@ timestamp < "${state.until}"`;
...
+                // so half the given interval and if interval is less than 15 sec then move interval by same.
                 if(err.code === API_THROTTLING_ERROR){
                     const nextPollInterval = state.poll_interval_sec < 10 ?
                         state.poll_interval_sec + 1:
                         10;
+                    const currentInterval = moment(state.until).diff(state.since, 'seconds');
+                    if (currentInterval <= 15) {
+                        const newStart = moment(state.since).add(currentInterval, 'seconds');
+                        state.since = newStart.toISOString();
+                        state.until = moment(newStart).add(15, 'seconds').toISOString();
```
kkuzmin, line 134: If `currentInterval <= 15 sec` we don't need to change `state.since` and `state.until`, but try to collect the interval again. Maybe put a warning message if we keep getting throttling error for the shortest interval (essentialy collection stopped.).

kkuzmin (changes requested):

kkuzmin (approved):

PR #250, opened 2022-01-20
[Paws Utility] : Collection start date gt than 30 days, then adjust the date to last 30 days

rakhimundhada15:
### Problem Description
DD show collection delay increase for couple of collector. Below is the one of the reason to increase the collection delay.
Customer can add the  any collection start date from UI, so seen some customer added the long back date(year ago). 
But ingest will accept only last 30 days data. We are unnecessary making the so many API call .

### Solution Description
If collection start date greater than 30 days, So adjust the start collection to last 30 days.

### DIFF
--- a/paws_utils.js
+++ b/paws_utils.js
@@ -17,7 +17,7 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
     const nowMoment = moment();
     // If collection start date gt than 30 days, ingest will not accept data, so adjust the date to last 30 days.
     const nextSinceMoment = curUntilMoment.isAfter(nowMoment) ?
-        nowMoment : nowMoment.diff(curUntilMoment, 'days') > 30 ? moment().subtract(30, 'days') : curUntilMoment;
+        nowMoment : nowMoment.diff(curUntilMoment, 'days') > 30 ? moment(nowMoment).subtract(30, 'days') : curUntilMoment;
     const daysDiff = nowMoment.diff(nextSinceMoment, 'days');
     const hoursDiff = nowMoment.diff(nextSinceMoment, 'hours');
     let nextUntilMoment;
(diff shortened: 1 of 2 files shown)


scott-jones-AlertLogic (approved):

paws_utils.js
```diff
@@ -15,8 +15,9 @@ function calcNextCollectionInterval(strategy, curUntilMoment, pollInterval) {
     const DELAY_IN_MINUTES = 15;
     const pollIntervalDelay = process.env.paws_poll_interval_delay && process.env.paws_poll_interval_delay <= 900 ? process.env.paws_poll_interval_delay : 300;
     const nowMoment = moment();
+    // If collection start date gt than 30 days, ingest will not accept data, so adjust the date to last 30 days.
     const nextSinceMoment = curUntilMoment.isAfter(nowMoment) ?
-        nowMoment : curUntilMoment;
+        nowMoment : nowMoment.diff(curUntilMoment, 'days') > 30 ? moment().subtract(30, 'days') : curUntilMoment;
```
kkuzmin, line 20: we should be able to use `nowMoment().subtract(30, 'days')`

kkuzmin (changes requested):

paws_utils.js
rakhimundhada15: If we used nowMoment() we get the nowMoment is not a function . If i try directly nowMoment.subtract(30, 'days') then nowMoment.diff(nextSinceMoment, 'days') return 0 days instead of 30 days.
So used moment() directly.

kkuzmin: right, that is because you need a copy of `nowMoment`, like `moment(nowMoment).subtract(30, 'days')`.
Otherwise `moment()` from line 17 and `moment()` from line 20 will be two different timestamps.

kkuzmin (changes requested):

kkuzmin (approved):

PR #251, opened 2022-01-31
[Googlestack driver] : Handle the GCP throttling error with max message size 4MB 

rakhimundhada15:
### Problem Description
Getting the Throttling error  with message size greater than 4 MB from GCP collector.
` Error: 8 RESOURCE_EXHAUSTED: Received message larger than max (4518352 vs. 4194304)`

### Solution Description
First reduce the interval into half and if not resolved for less than 15 sec then reduce the page size to half and pull the data.

### DIFF
--- a/collectors/googlestackdriver/collector.js
+++ b/collectors/googlestackdriver/collector.js
@@ -19,6 +19,7 @@ const packageJson = require('./package.json');
 
 const API_THROTTLING_ERROR = 8;
 const MAX_POLL_INTERVAL = 900;
+const MAX_PAGE_SIZE = 1000;
 
 const typeIdPaths = [
     {path: ['jsonPayload', 'fields', 'event_type', 'stringValue']},
@@ -99,7 +100,7 @@ timestamp < "${state.until}"`;
             }
         };
 
-        const pageSize = state.pageSize > 0 ? state.pageSize : 1000;
+        const pageSize = state.pageSize > 0 ? state.pageSize : MAX_PAGE_SIZE;
         let params = state.nextPage ?
             state.nextPage:
             {
@@ -161,10 +162,10 @@ timestamp < "${state.until}"`;
 
     _getNextCollectionState(curState, nextPage) {
         // Reset the page size for next collection as log collection completed for throttling interval
-        if (nextPage && nextPage.pageSize && nextPage.pageSize < 1000) {
-            nextPage.pageSize = 1000;
-        } else if (curState.pageSize && curState.pageSize < 1000) {
-            curState.pageSize = 1000;
+        if (nextPage && nextPage.pageSize && nextPage.pageSize < MAX_PAGE_SIZE) {
+            nextPage.pageSize = MAX_PAGE_SIZE;
+        } else if (curState.pageSize && curState.pageSize < MAX_PAGE_SIZE) {
+            curState.pageSize = MAX_PAGE_SIZE;
         }
         const {stream} = curState;
 
--- a/collectors/googlestackdriver/collector.js
+++ b/collectors/googlestackdriver/collector.js
@@ -18,6 +18,7 @@ const logging = require('@google-cloud/logging');
 const packageJson = require('./package.json');
 
 const API_THROTTLING_ERROR = 8;
+const MAX_POLL_INTERVAL = 900;
 
 const typeIdPaths = [
     {path: ['jsonPayload', 'fields', 'event_type', 'stringValue']},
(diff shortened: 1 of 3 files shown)


collectors/googlestackdriver/collector.js
```diff
@@ -119,16 +121,24 @@ timestamp < "${state.until}"`;
...
                 if(err.code === API_THROTTLING_ERROR){
-                    const nextPollInterval = state.poll_interval_sec < 10 ?
-                        state.poll_interval_sec + 1:
-                        10;
+                    const interval = state.poll_interval_sec < 60 ? 60 : state.poll_interval_sec;
+                    const nextPollInterval = state.poll_interval_sec < MAX_POLL_INTERVAL ?
+                        interval + 60 : MAX_POLL_INTERVAL;
                     const currentInterval = moment(state.until).diff(state.since, 'seconds');
-                    if (currentInterval <= 15) {
+                    if (currentInterval <= 15 && err.details.includes('Received message larger than max')) {
```
kkuzmin, line 134: Is it possible to use some error code instead of sub string search `err.details.includes('Received message larger than max')`? At some point Google can change the error message text.

collectors/googlestackdriver/collector.js
```diff
@@ -150,6 +160,12 @@ timestamp < "${state.until}"`;
     }
 
     _getNextCollectionState(curState, nextPage) {
+        // Reset the page size for next collection as log collection completed for throttling interval
+        if (nextPage && nextPage.pageSize && nextPage.pageSize < 1000) {
```
kkuzmin, line 164: please use a `MAX_PAGE_SIZE` constant

collectors/googlestackdriver/test/test.js
```diff
@@ -351,6 +356,25 @@ describe('Unit Tests', function() {
...
+                var collector = new GooglestackdriverCollector(ctx, creds);
+                const curState = {
+                    since: startDate.toISOString(),
+                    until: startDate.add(collector.pollInterval, 'seconds').toISOString(),
+                    poll_interval_sec: 1,
+                    stream: 'projects/test',
+                };
+                const nextPage = {"filter":"timestamp >= \"2022-01-21T00:00:15.000Z\"\ntimestamp < \"2022-01-22T00:00:15.000Z\"","pageSize":500,"resourceNames":["projects/test"],"pageToken":"EAA46o"};
+                const newState = collector._getNextCollectionState(curState, nextPage);
+                console.log('newState',newState);
```
kkuzmin, line 371: please clean up.

kkuzmin (changes requested):

collectors/googlestackdriver/collector.js
```diff
@@ -119,16 +121,24 @@ timestamp < "${state.until}"`;
...
+                        interval + 60 : MAX_POLL_INTERVAL;
                     const currentInterval = moment(state.until).diff(state.since, 'seconds');
-                    if (currentInterval <= 15) {
+                    if (currentInterval <= 15 && err.details.includes('Received message larger than max')) {
+                        // Reduce the page size to half to pull the data for throttling interval
+                        if (state.nextPage && state.nextPage.pageSize) {
+                            state.nextPage.pageSize = Math.ceil(state.nextPage.pageSize / 2);
+                        }
+                        else {
+                            state.pageSize = Math.ceil(params.pageSize / 2)
```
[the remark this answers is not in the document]
rakhimundhada15: No, If we get nextPage from APIs then it[ already conatin](https://github.com/alertlogic/paws-collector/blob/master/collectors/googlestackdriver/collector.js#L101) the filter , resourceName  and pageSize, I am just modifing the pageSize.  If nextpage not available then we need to [set](https://github.com/alertlogic/paws-collector/pull/251/files#diff-4ceb28ee2558e106d3a162c0641167b0a783ecdd9b01d487a614cc2685d21511R107) filter and resourceName and pageSize before making API call.

collectors/googlestackdriver/collector.js
rakhimundhada15: We  already checking error code [here](url). We get two different message for throttling .
1. `{code: 8 ,details : "Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute' of service 'logging.googleapis.com'" }`-> It resolved by just making the api call after 1 min or so.
2. `{code: 8 ,details : Received message larger than max (4776477 vs. 4194304)} `-> In this scenario we are reducing the page size to handle the error.

kkuzmin (approved):

PR #256, opened 2022-03-04
Changes for o365 node 14 update

sameerbag:
### Problem Description
Upgrade O365 to node 14

### Solution Description
Made node 14 changes and also bumped the package


IMRLN (approved):

collectors/o365/local/run-sam.sh
```diff
@@ -7,7 +7,7 @@ SRC_SAM_TEMPLATE="${SCRIPT_DIR}/sam-template.yaml"
 SRC_ENV_FILE="${SCRIPT_DIR}/${ENV_FILE_NAME}"
 SRC_EVENT_FILE="${SCRIPT_DIR}/events/${EVENT_FILE_NAME}"
 RUN_DIR=${SCRIPT_DIR}/../
-
+PROFILE_NAME=""
```
kkuzmin, line 10: let's use `default` profile instead of  `""`.

kkuzmin (approved):
One minor suggestion

collectors/o365/local/run-sam.sh
sameerbag: Done

PR #275, opened 2022-04-07
Changes for sentinel one node 14 update

sameerbag:
### Problem Description
Sentinel one collector update to node 14

### Solution Description
Changes for sentinel one collector update to node 14


IMRLN (approved):

kkuzmin (changes requested):
Please update local invocations config files. like https://github.com/alertlogic/paws-collector/pull/274/files

kkuzmin (approved):

PR #289, opened 2022-06-17
Async Version fix

imkariya:
### Problem Description
Async Version fix

### Solution Description

Updated package json in following collectors

- auth0                   
- ciscoduo                
- googlestackdriver       
- o365                    
- sentinelone     
- carbonblack             
- gsuite                  
- okta                    
- sophos
- ciscoamp                
- crowdstrike             
- mimecast                
- salesforce              
- sophossiem

### DIFF
--- a/collectors/auth0/package.json
+++ b/collectors/auth0/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth0-collector",
-  "version": "1.1.31",
+  "version": "1.1.32",
   "description": "Alert Logic AWS based Auth0 Log Collector extension",
   "repository": {},
   "private": true,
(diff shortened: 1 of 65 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "@alertlogic/al-aws-collector-js": "4.1.2",
     "datadog-lambda-js": "2.23.0",
-    "async": "3.1.0",
+    "async": "3.2.2",
```
kkuzmin, line 37: we need to make paws-collector/package.json change as a separate PR. And once it is shipped we need to bump paws-collector dependency version in all collectors.

kkuzmin (changes requested):

collectors/auth0/package.json
```diff
@@ -22,7 +22,7 @@
     "auth0": "2.27.1",
     "@alertlogic/al-collector-js": "^2.0.4",
     "@alertlogic/paws-collector": "^2.1.2",
```
kkuzmin, line 24: please update it to 2.1.4

kkuzmin (approved):

PR #291, opened 2022-06-21
Gsuit moment version fix

imkariya:
### Solution Description

- Deprecated moment version fix in package.json


kkuzmin (changes requested):
please update package version.

PR #300, opened 2022-08-11
[Cisco AMP]: Handle throttling error

rakhimundhada15:
### Problem Description
Collector try to pull the data for one month interval and showing throttling error continuously .Throttling error not  resolved after couple of time tried .
`CAMP000001 Collecting data for Events from 2022-07-01T10:46:51.950Z till 2022-07-31T18:18:28.542Z`
`CAMP000003 API hourly Limit Exceeded. The quota will be reset at 2022-08-11T07:42:19.690Z"`

### Solution Description
Used the day cap if start time is greater than 24 hr and reduce the time interval and try to fetch if not then reduce the time interval again till 60 sec.

### DIFF
--- a/collectors/ciscoamp/collector.js
+++ b/collectors/ciscoamp/collector.js
@@ -107,51 +107,51 @@ class CiscoampCollector extends PawsCollector {
         }
         else {
 
-            utils.getAPILogs(baseUrl, base64EncodedString, apiUrl, state, [], process.env.paws_max_pages_per_invocation)
-                .then(({ accumulator, nextPage, resetSeconds, totalLogsCount, discardFlag }) => {
-                    if (resetSeconds) {
-                        const extraBufferSeconds = 60;
-                        resetSeconds = resetSeconds + extraBufferSeconds;
-                        state.apiQuotaResetDate = moment().add(resetSeconds, "seconds").toISOString();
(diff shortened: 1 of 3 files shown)


collectors/ciscoamp/collector.js
```diff
@@ -96,6 +96,11 @@ class CiscoampCollector extends PawsCollector {
         if (state.apiQuotaResetDate && moment().isBefore(state.apiQuotaResetDate)) {
             AlLogger.info(`CAMP000002 API hourly Limit Exceeded. The quota will be reset at ${state.apiQuotaResetDate}`);
             state.poll_interval_sec = 900;
```
kkuzmin, line 98: I wonder if we hit API hourly limit does it actually worth trying to make another call in 15 mins. Looks like we will hit throttling issue again.

kkuzmin (changes requested):
Please update package.json version

collectors/ciscoamp/collector.js
rakhimundhada15: Api return the [x-ratelimit-remaining](https://github.com/alertlogic/paws-collector/blob/master/collectors/ciscoamp/utils.js#L27) and  when it reset (x-ratelimit-reset) ,accordingly we are setting [apiQuotaResetDate](https://github.com/alertlogic/paws-collector/blob/master/collectors/ciscoamp/collector.js#L110) . There is no fix time after which we have to call.

kkuzmin (approved):

PR #301, opened 2022-08-18
Increase the poll_interval to avoid multiple api call

rakhimundhada15:
### Problem Description
Cisco duo api making call in every sec for Telephony,Administrator and for offlineEnrollment stream.  Increasing the AWS costing.
### Solution Description
Seems  like thee is not much data for above streams, So Increase the poll interval to 60 if next page available else increase by 15 min incase time interval is less than 120 min.


collectors/ciscoduo/collector.js
```diff
@@ -178,10 +182,11 @@ class CiscoduoCollector extends PawsCollector {
             };
         } else {
             //There is no next page concept for this API, So Setting up the next state mintime using the last log (Unix timestamp + 1).
+            // call after 60 sec to avoid multiple api call
```
kkuzmin, line 185: I think the actual issue is [here](https://github.com/alertlogic/paws-collector/blob/aa2343b20b9e208393ca11de3d1fa20ad7ee4b32/collectors/ciscoduo/utils.js#L42-L46)
Please check if we can avoid using pagination for continuous time interval polling.
Pagination is a way to get all the data which collector queries within one collection interval. And collection intervals should be moved separately from pagination.

kkuzmin (changes requested):

collectors/ciscoduo/collector.js
```diff
@@ -154,15 +154,24 @@ class CiscoduoCollector extends PawsCollector {
...
-            const { nextUntilMoment, nextSinceMoment, nextPollInterval } = calcNextCollectionInterval('no-cap', untilMoment, this.pollInterval);
-
-            return {
-                stream: curState.stream,
-                mintime: nextSinceMoment.unix(),
-                poll_interval_sec: nextPollInterval
-            };
-
+            if (curState.mintime) {
+                // If next page is null check min time stamp is less than 1 hr move by 1 min else set the last hour time stamp as newMintime
```
kkuzmin, line 158: please fix the comment. We do not check for `page is null`

collectors/ciscoduo/collector.js
```diff
@@ -154,15 +154,24 @@ class CiscoduoCollector extends PawsCollector {
...
+                const lastHourMoment = moment().subtract(1, 'hours').unix();
```
kkuzmin, line 159: I think we can use `untilMoment` variable here.

collectors/ciscoduo/collector.js
```diff
@@ -154,15 +154,24 @@ class CiscoduoCollector extends PawsCollector {
...
-
-            const { nextUntilMoment, nextSinceMoment, nextPollInterval } = calcNextCollectionInterval('no-cap', untilMoment, this.pollInterval);
-
-            return {
-                stream: curState.stream,
-                mintime: nextSinceMoment.unix(),
-                poll_interval_sec: nextPollInterval
-            };
-
+            if (curState.mintime) {
```
kkuzmin, line 157: I guess `curState.mintime` is always set and this `if` doesn't make a lot of sense.

collectors/ciscoduo/collector.js
```diff
@@ -154,15 +154,24 @@ class CiscoduoCollector extends PawsCollector {
...
+                // If next page is null check min time stamp is less than 1 hr move by 1 min else set the last hour time stamp as newMintime
+                const lastHourMoment = moment().subtract(1, 'hours').unix();
+                const newMintime = Math.max(curState.mintime + 1, lastHourMoment);
```
kkuzmin, line 160: IIRC `mintime` is in seconds.

collectors/ciscoduo/collector.js
```diff
@@ -154,15 +154,24 @@ class CiscoduoCollector extends PawsCollector {
...
+                return {
+                    stream: curState.stream,
+                    mintime: newMintime,
+                    poll_interval_sec: 60
```
kkuzmin, line 164: I think if collector catches up with backlog and startc collecting last hour data then we can increase pool interval to 15 mins.

kkuzmin (changes requested):

kkuzmin (approved):

PR #302, opened 2022-09-29
CiscoAMP collector[1.0.34] : Refactor the code for Event stream and Throttle Error

rakhimundhada15:
### Problem Description

1. Customer having CiscoAMP collector and its Event stream not moving collection windows(lacking behind almost 1 and half month)

Root Cause :

-  When we make the api call for particular timestamp(`Events from 2022-07-30T00:00:00.000Z till 2022-07-30T00:00:57.000Z`),every time total log counts are different for same time interval. As per previous code we are discarding the logs if it not match with previous count stored in state object. So collection window not get updated .

 "results": {
            **"total": 131625,**
            "current_item_count": 500,
            "index": 500,
            "items_per_page": 500
        }

### Solution Description

1. Cisco AMP API only have startDate (https://api.amp.cisco.com/v1/events?**start_date**=2022-09-19T00%3A00%3A00.000Z) and not end time stand, so total count will vary. APIs return the data in descending order by default . Handle how to start next collection as below

- If Api return next link value in metadata use for next collection .
- Else pick the date from first logs message and set new start value.

"links": {
            "self": "https://api.amp.cisco.com/v1/events?start_date=2022-09-19T00%3A00%3A00.000Z&offset=500",
            "prev": "https://api.amp.cisco.com/v1/events?start_date=2022-09-19T00%3A00%3A00.000Z&offset=0",
            "next": "https://api.amp.cisco.com/v1/events?start_date=2022-09-19T00%3A00%3A00.000Z&offset=1000"
        },

2. Also Handle the RateLimit error properly.

### DIFF
--- a/collectors/ciscoamp/utils.js
+++ b/collectors/ciscoamp/utils.js
@@ -32,8 +32,11 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
                     }
                     else {
                         if (state.stream === Events && accumulator.length > 0) {
-                            //Api return the data in desending order, so set new start date fom first record.
-                            newSince = accumulator[0].date;
+                            // Api return the data in desending order, so set new start date form first record if not available pull date from fallup records.
+                            newSince = accumulator[0].date ? accumulator[0].date : accumulator[1].date ? accumulator[1].date : accumulator[2].date;
+                            if (!newSince) {
+                                reject(`CAMP000005 Date is not available in Events api response`);
+                            }
                         }
                         resolve({ accumulator, nextPage: undefined, newSince });
                     }
--- a/collectors/ciscoamp/utils.js
+++ b/collectors/ciscoamp/utils.js
@@ -7,9 +7,7 @@ const Events = 'Events';
 function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPagesPerInvocation) {
     let pageCount = 0;
     let nextPage;
-    let resetSeconds = null;
-    let totalLogsCount = state.totalLogsCount;
-    let discardFlag = false;
+    let newSince = null;
 
     let restServiceClient = new RestServiceClient(baseUrl);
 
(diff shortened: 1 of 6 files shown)


collectors/ciscoamp/utils.js
```diff
@@ -22,39 +20,30 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
...
-                            apiUrl = url.parse(body.metadata.links.next).path;
-                            getData();
-                        }
-                        else {
-                            resolve({ accumulator, nextPage, resetSeconds, totalLogsCount, discardFlag });
+                        if (state.stream === Events && accumulator.length > 0) {
+                            //Api return the data in desending order, so set new start date fom first record.
+                            newSince = accumulator[0].date;
                         }
+                        resolve({ accumulator, nextPage, newSince });
```
kkuzmin, line 38: I think we need to pass `nextPage` as `undefined` here because we iterated  through all the pages. Otherwise `nextPage` value will be the last page link.

collectors/ciscoamp/utils.js
```diff
@@ -22,39 +20,30 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
...
                     }
                 }).catch(err => {
                     reject(err);
                 });
             }
             else {
                 nextPage = apiUrl;
-                resolve({ accumulator, nextPage, resetSeconds, totalLogsCount, discardFlag });
+                resolve({ accumulator, nextPage, newSince });
```
kkuzmin, line 46: I think we need to explicitly set `newSinse` to `undefined` in order to indicate that we haven't iterated through all the pages within current invocation.

kkuzmin (changes requested):

collectors/ciscoamp/utils.js
rakhimundhada15: Here newSince value is by default null. But still i will explicitly set to undefined.

collectors/ciscoamp/utils.js
```diff
@@ -22,39 +20,30 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
...
-                        }
-                        if (body.metadata.links.next) {
-                            apiUrl = url.parse(body.metadata.links.next).path;
-                            getData();
-                        }
-                        else {
-                            resolve({ accumulator, nextPage, resetSeconds, totalLogsCount, discardFlag });
+                        if (state.stream === Events && accumulator.length > 0) {
+                            //Api return the data in desending order, so set new start date fom first record.
+                            newSince = accumulator[0].date;
```
kkuzmin, line 36: maybe check if `date` field exists. And if not output an error and fall back to using `accumulator[1]`, `accumulator[2]`, etc.

kkuzmin (approved):

PR #303, opened 2022-09-29
[Cisco Duo collector] Increase the poll Interval secs from 1 to 60 sec

rakhimundhada15:
### Problem Description
Customer still able to see a couple of Rate limit error in every 15 min.
{
    "code": 42901,
    "level": "warn",
    "message": "CDUO000003 API Request Limit Exceeded Too Many Requests",
    "requestId": "8113cfaf-1017-5472-934f-94007e9d9c91",
    "stat": "FAIL",
    "timestamp": "2022-09-29T09:59:48.937Z"
}

RootCouse:
If collector is lacking behind more than 15 min it will make a api call every secs.

### Solution Description
AS per [doc](https://duo.com/docs/adminapi#authentication-logs)(`We recommend requesting logs no more than once per minute`.)  i.e make api call one pr minutes .
So updated the poll interval sec to 60sec from 1sec.

### DIFF
--- a/collectors/ciscoduo/collector.js
+++ b/collectors/ciscoduo/collector.js
@@ -140,13 +140,13 @@ class CiscoduoCollector extends PawsCollector {
             const untilMoment = moment(parseInt(curState.maxtime));
              // Used hour-cap instead of making api call for 1 min interval, may help to reduce throtling issue.
             const { nextUntilMoment, nextSinceMoment, nextPollInterval } = calcNextCollectionInterval('hour-cap', untilMoment, this.pollInterval);
-
+            const nextPollIntervalSec = nextPollInterval > 1 ? nextPollInterval : 60;
             return {
                 stream: curState.stream,
                 mintime: nextSinceMoment.valueOf(),
                 maxtime: nextUntilMoment.valueOf(),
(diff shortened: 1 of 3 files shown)


collectors/ciscoduo/collector.js
```diff
@@ -140,13 +140,13 @@ class CiscoduoCollector extends PawsCollector {
             const untilMoment = moment(parseInt(curState.maxtime));
              // Used hour-cap instead of making api call for 1 min interval, may help to reduce throtling issue.
             const { nextUntilMoment, nextSinceMoment, nextPollInterval } = calcNextCollectionInterval('hour-cap', untilMoment, this.pollInterval);
-
+            const nextPollIntervalSec = nextPollInterval > 1 ? nextPollInterval : 60;
```
kkuzmin, line 143: > We recommend requesting logs no more than once per minute.

Sounds like we need to have this:
```
const nextPollIntervalSec = nextPollInterval > 60 ? nextPollInterval : 60;
```

Meaning don't have a intervals less than 1 minute.

kkuzmin (changes requested):

kkuzmin (approved):

PR #304, opened 2022-10-07
Cisco AMP Collector : Fix the duplication of message Scenario 

rakhimundhada15:


collectors/ciscoamp/collector.js
```diff
@@ -96,8 +96,8 @@ class CiscoampCollector extends PawsCollector {
         utils.getAPILogs(baseUrl, base64EncodedString, apiUrl, state, [], process.env.paws_max_pages_per_invocation)
             .then(({ accumulator, nextPage, newSince }) => {
                 state.apiQuotaResetDate = null;
-                if (!nextPage && state.stream == Events && newSince) {
-                    state.until = moment(newSince).toISOString();
```
kkuzmin, line 100: Do we do `.toISOString()` conversion somewhere else before it gets serialised into SQS message?

kkuzmin (changes requested):

collectors/ciscoamp/collector.js
rakhimundhada15: Yes i am doing here- https://github.com/alertlogic/paws-collector/pull/304/files#diff-6afff9141604ac4282cb8f121376b92e9886b7225dd70f56ba3b316dec3f22dbR34

collectors/ciscoamp/utils.js
```diff
@@ -32,11 +33,7 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
                     }
                     else {
                         if (state.stream === Events && accumulator.length > 0) {
-                            // Api return the data in desending order, so set new start date form first record if not available pull date from fallup records.
-                            newSince = accumulator[0].date ? accumulator[0].date : accumulator[1].date ? accumulator[1].date : accumulator[2].date;
```
kkuzmin, line 36: I think we need to keep this.

kkuzmin (changes requested):

collectors/ciscoamp/utils.js
rakhimundhada15: if there is no date available we can take from Until value from state object

kkuzmin: this code was added to make sure we find `date` property in most recently collected messages. This was done to make sure `newSince` moves, otherwise it'll result in recollections. 
I think it should be
```
// Api return the data in desending order, so set new start date form first record if not available pull date from fallup records.
  findNewSince = accumulator[0].date ? accumulator[0].date : accumulator[1].date ? accumulator[1].date : accumulator[2].date;
  if (!findNewSince) {
      reject(`CAMP000005 Date is not available in Events api response`);
  }
  newSince = moment(findNewSince).add(1, 'seconds').toISOString();
```

kkuzmin (approved):

PR #305, opened 2022-10-11
Cisco AMP collector : Stored the firstPage date in state until and used it for next collection once all pages collection done.

rakhimundhada15:
### Problem Description
Added duplication of message. 
@Scenario : When collector is lacking behind and pages are more than max Invocation(default is 10).
Cisco AMP api return the data in descending order, as we go through pages , data become older. If we pick the date from last received page for next collection then that date is older and not the current value still which we all ready fetch data. 
When we start next collection it adding duplicate data.
Explain  with example :
If I make api call(https://api.amp.cisco.com/v1/events?start_date=2022-10-09T13:49:45.000Z) at 11-10-2022T05:52:000Z (UTC) ,api return the data from last min to start_date value .
image

Total record is "total": 5436,  each page it return 500 records. In one invocation we read max 10 pages. In this scenario we need one more invocation to fetch rest of data.
When we read all pages and reached to last one then metadata data does not have **next page link** and so we were fetching the date from received message which is older as show below in screen shot
image

### Solution Description
Stored the date of first message from first page in state.until .Used state until for next collection once all pages data we collect.


collectors/ciscoamp/utils.js
```diff
@@ -33,13 +33,9 @@ function getAPILogs(baseUrl, authorization, apiUrl, state, accumulator, maxPages
...
-                            // Api return the data in desending order, so set new start date form first record if not available pull date from fallup records.
-                            const findDate = accumulator[0].date ? accumulator[0].date : accumulator[1].date ? accumulator[1].date : accumulator[2].date;
-                            if (findDate) {
-                                newSince = moment(findDate).add(1, 'seconds');
-                            } else {
-                                reject(`CAMP000005 Date is not available in Events api response`);
-                            }
+                            // Api return the data in desending order, If there are lots of pages, when we go through page it return older data and old date. 
+                            // As we already read the data till current time stamp, so adding 1 sec for next collection to avoid duplication of records.
+                            newSince = moment(state.until).add(1, 'seconds');
```
imranalisyed506, line 38: LGTM +1

imranalisyed506 (approved):
LGTM +1

kkuzmin (approved):

PR #310, opened 2022-11-17
Add collector id as Hostname in Okta collector

kkuzmin:
### Solution Description
Include `hostname` as `collector_id` into log messages in order to make incident deployment resolution work.


rakhimundhada15 (approved):
LGTM

PR #313, opened 2022-11-28
Configure the DLQ in SQS for new deployed collectors

rakhimundhada15:
### Problem Description
Few collector facing collection delay because ingest service returning 400-'body encoding invalid'.

### Solution Description

1. Create DLQ in each region using shared resource template -https://algithub.pd.alertlogic.net/defender/paws-collectors-deploy-pipeline/pull/410

2. Map the shared dead-letter queue ARN in azcollect and set the value of ExistingDeadLetterQueueArn - https://algithub.pd.alertlogic.net/defender/azcollect/pull/257

3. Added below parameter to differentiate the DLQ message and identify the message from which collector.
- lambda_function_name
- customer_id
- account_id

4. Configured the existing DLQ for each collector using CFN template while deploying the new collector. 

5. Move the failed message to DLQ to avoid collection delay and we can investigate the data later to improve if any formatting or other issue found.

### DIFF
--- a/cfn/paws-collector-shared.template
+++ b/cfn/paws-collector-shared.template
@@ -167,6 +167,10 @@
         "ExistingControlSnsArn" :{
             "Description": "SNS topic ARN used for triggering collector checkins an updates",
             "Type": "String"
+        },
+        "ExistingDeadLetterQueueArn":{
+            "Description": "Dead-letter queue to collect failed message",
+            "Type": "String"
         }
     },
     "Resources":{
@@ -248,20 +252,7 @@
             "MessageRetentionPeriod": 1209600,
             "RedrivePolicy": {
                 "deadLetterTargetArn": {
-                    "Fn::Join": [
-                        "",
-                        [
-                            "arn:aws:sqs:",
-                            {
-                                "Ref": "AWS::Region"
-                            },
-                            ":",
-                            {
-                                "Ref": "AWS::AccountId"
-                            },
-                            ":Paws-shared-deadLetterQueue"
-                        ]
-                    ]
+                    "Ref": "ExistingDeadLetterQueueArn"
                 },
                 "maxReceiveCount": 10
             },
--- a/cfn/paws-collector-shared.template
+++ b/cfn/paws-collector-shared.template
@@ -254,7 +254,7 @@
                 "deadLetterTargetArn": {
                     "Ref": "ExistingDeadLetterQueueArn"
                 },
-                "maxReceiveCount": 10
+                "maxReceiveCount": 3
             },
             "Tags": [
                 {
--- a/cfn/paws-collector-shared.template
+++ b/cfn/paws-collector-shared.template
@@ -246,6 +246,25 @@
          "Properties":{
             "VisibilityTimeout": 900,
             "MessageRetentionPeriod": 1209600,
+            "RedrivePolicy": {
+                "deadLetterTargetArn": {
+                    "Fn::Join": [
+                        "",
+                        [
+                            "arn:aws:sqs:",
+                            {
+                                "Ref": "AWS::Region"
+                            },
+                            ":",
+                            {
+                                "Ref": "AWS::AccountId"
+                            },
+                            ":Paws-shared-deadLetterQueue"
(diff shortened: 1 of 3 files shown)


cfn/paws-collector.template
```diff
@@ -685,6 +685,25 @@
...
+                        [
+                            "arn:aws:sqs:",
+                            {
+                                "Ref": "AWS::Region"
+                            },
+                            ":",
+                            {
+                                "Ref": "AWS::AccountId"
+                            },
+                            ":Paws-shared-deadLetterQueue"
```
kkuzmin, line 701: THe DLQ should exist prior to using it here. Please update the https://algithub.pd.alertlogic.net/defender/paws-collectors-deploy-pipeline/blob/master/cfn/paws-shared-resources.template

kkuzmin: If we use same DLQ for all collectors can we differentiate DLQ message to identify which collector a particular DLQ message belongs to?

kkuzmin (changes requested):

cfn/paws-collector.template
rakhimundhada15: Added extra param to identify the message from which collector in case it failed and went in DLQ.

paws_collector.js
```diff
@@ -686,6 +686,13 @@ class PawsCollector extends AlAwsCollector {
     };
 
     _storeCollectionState(pawsState, privCollectorState, invocationTimeout, callback) {
+        // Added extra parameter to identify the message from which collector in case it failed and went into DLQ
+        let extraParam = {
+            lambda_function_name: process.env.AWS_LAMBDA_FUNCTION_NAME,
+            customer_id: process.env.customer_id,
+            account_id: this._awsAccountId
```
kkuzmin, line 693: I think we don't need `_awsAccountId` here.
Please add `collector_id` instead.

paws_collector.js
```diff
@@ -686,6 +686,13 @@ class PawsCollector extends AlAwsCollector {
     };
 
     _storeCollectionState(pawsState, privCollectorState, invocationTimeout, callback) {
+        // Added extra parameter to identify the message from which collector in case it failed and went into DLQ
+        let extraParam = {
+            lambda_function_name: process.env.AWS_LAMBDA_FUNCTION_NAME,
```
kkuzmin, line 691: Don't think we need function name either

cfn/paws-collector.template
```diff
@@ -685,6 +690,12 @@
          "Properties":{
             "VisibilityTimeout": 900,
             "MessageRetentionPeriod": 1209600,
+            "RedrivePolicy": {
+                "deadLetterTargetArn": {
+                    "Ref": "ExistingDeadLetterQueueArn"
+                },
+                "maxReceiveCount": 10
```
kkuzmin, line 697: let's start with 3

cfn/paws-collector.template
```diff
@@ -144,6 +144,11 @@
             "Type": "String",
             "Default": "true",
             "AllowedValues":["true"]
+        },
+        "ExistingDeadLetterQueueArn":{
```
kkuzmin, line 148: this template is intended to create all the resources. In this case create DLQ sqs and reference it's ARN in `PawsPollStateQueue`
I think for the sake of time you can just add a TODO item to add DLQ into this template because it doesn't look like anyone is going to use it.

cfn/paws-collector-shared.template
```diff
@@ -246,6 +250,12 @@
          "Properties":{
             "VisibilityTimeout": 900,
             "MessageRetentionPeriod": 1209600,
+            "RedrivePolicy": {
+                "deadLetterTargetArn": {
+                    "Ref": "ExistingDeadLetterQueueArn"
+                },
+                "maxReceiveCount": 10
```
kkuzmin, line 257: let's start with 3

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: Okay

PR #314, opened 2022-12-15
Extract the error code from ingest message and send  to PawsIngestApi metrics

rakhimundhada15:
Currently PawsIngestApi metrics showing  errorCode : 'unknown`  for All PAWS collectors as al-aws-collector.js  just returning [exception.message](https://github.com/alertlogic/al-aws-collector-js/blob/master/al_aws_collector.js#L68) and not whole object. 
Reference pr -https://github.com/alertlogic/al-aws-collector-js/pull/74
### Solution Description
 Extracting the error code from ingest message and sending to PawsIngestApi . It help to check which all ingest error collector getting.


paws_collector.js
```diff
@@ -549,7 +549,14 @@ class PawsCollector extends AlAwsCollector {
         Promise.all(promises).then((res) => {
             return callback(null, privCollectorState, nextInvocationTimeout);
         }).catch((err) => {
-            collector.reportErrorToIngestApi(err, () => {
+            let errorObject = {};
+            if (typeof (err) === 'object') {
+                errorObject = Object.assign(errorObject, err);
+            } else {
+                let splitErrorMessage = err.split(':');
```
kkuzmin, line 556: I think we need a sanity check for `err` is of a `String` type.

paws_collector.js
```diff
@@ -549,7 +549,14 @@ class PawsCollector extends AlAwsCollector {
...
+                errorObject.errorCode = splitErrorMessage[1].slice(1, 4);
```
kkuzmin, line 557: Please add a test to this `else` branch

kkuzmin: I think we also need to check that `errorCode` can be converted to a number, like `300 < N < 600`. Otherwise it may contain some garbage.

kkuzmin: Please have a look if we already have an error parsing function. I suggest we either extract this code into a new function or merge it with existing function.

kkuzmin (changes requested):

paws_collector.js
```diff
@@ -549,12 +549,30 @@ class PawsCollector extends AlAwsCollector {
...
-            collector.reportErrorToIngestApi(err, () => {
-                return callback(err);
+            let error = collector.parseError(err);
+            collector.reportErrorToIngestApi(error, () => {
+                return callback(error);
             });
         });
     }
 
+    parseError(error) {
```
kkuzmin, line 559: it sounds like this function belongs to `al-collector-js` because same function might be helpful for Azure collectors as well.

kkuzmin: Let's maybe move it to `https://github.com/alertlogic/al-collector-js/blob/npm/al_util.js` or create a new `al_error.js` module for all error handling functions.

kkuzmin: Or IMHO a better idea is to return error objects with `errorCode` property from `al-aws-collector-js`. For example change this https://github.com/alertlogic/al-aws-collector-js/blob/master/al_aws_collector.js#L681
to return 
```
{
  errorCode: AWSC0018,
  errorMesasges: "AWSC0018 failed to send the logmsgs : ${exception.message}"
  httpErrorCode: 404
}
```

And amend `al-aws-collector-js` lib to extract HTTP error codes and pass them as separate field.

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: move the code in al-aws-collector and also return error object from it. Reference [pr](https://github.com/alertlogic/al-aws-collector-js/pull/88)

rakhimundhada15:
closing this pr as this change move in al-aws-collector.js lib. Reference [pr](https://github.com/alertlogic/al-aws-collector-js/pull/88). Raise new pr to use latest version of al-aws-collector .

PR #323, opened 2023-02-03
Converted the time from second to milliseconds to get correct date and collection delay secs

rakhimundhada15:
Cisco duo collector use epochs unix time stamp ,so converted it from seconds to milliseconds to get correct date and collection delay second.

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -674,7 +674,7 @@ class PawsCollector extends AlAwsCollector {
     reportCollectionDelay(lastCollectedTs, callback) {
         const nowMoment = moment();
         // converting the time from seconds to milliseconds . 
-        const lastCollectedMoment = !isNaN(lastCollectedTs) && lastCollectedTs.toString().length < 13 ? moment(parseInt(lastCollectedTs * 1000))  :  moment(lastCollectedTs);
+        const lastCollectedMoment = !isNaN(lastCollectedTs) && lastCollectedTs.toString().length == 10 ? moment.unix(lastCollectedTs) : moment(lastCollectedTs);
         const delayDuration = moment.duration(nowMoment.diff(lastCollectedMoment));
         const collectionDelaySec = Math.floor(delayDuration.asSeconds());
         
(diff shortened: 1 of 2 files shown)


paws_collector.js
```diff
@@ -673,7 +673,8 @@ class PawsCollector extends AlAwsCollector {
     
     reportCollectionDelay(lastCollectedTs, callback) {
         const nowMoment = moment();
-        const lastCollectedMoment = moment(lastCollectedTs);
+        // converting the time from seconds to milliseconds . 
+        const lastCollectedMoment = !isNaN(lastCollectedTs) && lastCollectedTs.toString().length < 13 ? moment(parseInt(lastCollectedTs * 1000))  :  moment(lastCollectedTs);
```
kkuzmin, line 677: I'd rather suggest we use seconds precision timestamps to calculate collection delay.
I don't think we need to measure collection delay at a milliseconds precision.

kkuzmin (changes requested):

paws_collector.js
rakhimundhada15: we are measuring collection delay base on seconds only as [here](https://github.com/alertlogic/paws-collector/pull/323/files#diff-82fd638370f7c86cddbb67ebf81bc2952ba84a1e7dc12ce1ed0fe1da30045492L678). 

Only i am converting seconds to milliseconds to get correct date . moment(lastCollectedTs) , for cisco duo  lastCollectdTs is unix() date so we need conversion. Other wise return wrong date 
ex. 
moment(1675435692) = Tue Jan 20 1970 09:23:55 GMT+0000

kkuzmin: you're right.
In this case I suggest we use `moment.unix()` to convert timestamp from Unix seconds, like
```
lastCollectedTs.toString().length == 10 ? moment.unix(lastCollectedTs) ?  moment(lastCollectedTs);
```

kkuzmin (approved):

PR #324, opened 2023-02-06
Update poll-Interval base on number of stream and update the state message.

rakhimundhada15:
### Problem Description

1. Few collector still facing throttle issue.
2. CiscoDuo collector not reporting collection delay metrics on DD

### Solution Description

- Cisco duo api allow 1 api call per min. So we increasing poll interval delay base on number of stream.
If there are 4 stream then each stream call after 4 min.

- Update state message and added since and until so that Paws lib will picked it value to calculate collection delay.


kkuzmin (commented):
@rakhimundhada15 please update `package.json`

kkuzmin (approved):

PR #330, opened 2023-05-09
[o365 Collector]: Handle the ingest error introduce due to deduplication changes

rakhimundhada15:
### Problem Description :
Collector get the data from api and stored the data in DDB before sending to ingest. if ingest send any error then collector tried the same , again after some time ; as data is already present in DDB ,so it mark all the data as duplicate. We are loosing data here.

### Solution Description

- Delete those entry from ddb which got error from ingest service while sending `/data/logmsgs`.
- Return the original error back to collector ,so that it will add same state in SQS message to retry again for fail state.
- Change the expire_ttl_hours to seconds ,so that it easy to set value in seconds from env variable rather than converting from hr to seconds.

### DIFF
--- a/paws_collector.js
+++ b/paws_collector.js
@@ -850,14 +850,11 @@ class PawsCollector extends AlAwsCollector {
         var promises = [];
         let duplicateCount = 0;
         logs.forEach(message => {
-            const cid = collector.cid ? collector.cid : 'none';
-            const collectorId = collector._collectorId;
-            const messageHash = collector.getHash(message);
-            const itemId = `${cid}_${collectorId}_${messageHash}`;
+            const itemId = collector.getDeDupItemId(message);
             const params = {
                 Item: {
                     Id: { S: itemId },
-                    CollectorId: { S: collectorId },
+                    CollectorId: { S: collector._collectorId },
                     OrigMessageId: { S: message[`${paramName}`] },
                     ExpireDate: { N: collector.calculateExpiryTs() }
                 },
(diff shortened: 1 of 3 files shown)


paws_collector.js
```diff
@@ -890,8 +897,77 @@ class PawsCollector extends AlAwsCollector {
...
+        let collector = this;
+        const cid = collector.cid ? collector.cid : 'none';
+        const collectorId = collector._collectorId;
+        const tableName = collector._pawsDeDupLogsTableName
+        const promises = [];
+        let currentBatch = [];
+
+        for (const message of logs) {
+            const messageHash = collector.getHash(message);
+            const itemId = `${cid}_${collectorId}_${messageHash}`;
```
scott-jones-AlertLogic, line 935: let's have a separate function to make the item ID so we don't replicate logic between here and `removeDuplicatedItem()`

paws_collector.js
```diff
@@ -890,8 +897,77 @@ class PawsCollector extends AlAwsCollector {
...
+            const key = {
+                Id: { S: itemId },
+                CollectorId: { S: collectorId }
+            }
+            currentBatch.push({ DeleteRequest: { Key: key } });
+            const itemSizeBytes = JSON.stringify(currentBatch).length
+
+            if (currentBatch.length === DDB_DELETE_BATCH_OPTIONS.maxBatchSize || currentBatch.length * itemSizeBytes >= DDB_DELETE_BATCH_OPTIONS.maxBatchSizeBytes) {
```
scott-jones-AlertLogic, line 943: please move this condition to a separate function like 

```suggestion
            if this._isFullBatch(currentBatch) {
                // ...
            }
```

scott-jones-AlertLogic (commented):
just some stylistic comments but this looks good to me

paws_collector.js
```diff
@@ -890,8 +894,95 @@ class PawsCollector extends AlAwsCollector {
...
+     * @param {*} message 
+     * @returns itemId
+     */
+    getDeDupItemId(message) {
+        const collector = this;
+        const cid = collector.cid ? collector.cid : 'none';
+        const collectorId = collector._collectorId;
+        const messageHash = collector.getHash(message);
+        const itemId = `${cid}_${collectorId}_${messageHash}`;
+        return itemId;
```
neil-bleasdale, line 962: Directly returning ```` `${cid}_${collectorId}_${messageHash}` ```` would probably be a bit clearer here I think

paws_collector.js
```diff
@@ -37,11 +37,16 @@ const ERROR_CODE_DUPLICATE_STATE = 'DUPLICATE_STATE';
...
+const DEDUP_LOG_TTL_SECONDS = 86400;
 const DDB_OPTIONS = {
     maxRetries: 10,
     ConsistentRead: true
 };
+
+const DDB_DELETE_BATCH_OPTIONS = {
+    maxBatchSize: 25,
+    maxBatchSizeBytes: 16 * 1024 * 1024
+}
```
neil-bleasdale, line 49: Are we missing a semi-colon here? Just for the sake of consistency and placating the linter

neil-bleasdale (approved):
A couple of really nit picking comments aside I think this looks great!

imranalisyed506 (approved):

PR #333, opened 2023-06-21
[ENG-43649 - Paws lib ] Send the status to collector_status service

rakhimundhada15:
### Problem Description
Stop sending status to ingest directly. [al-aws-collector pr](https://github.com/alertlogic/al-aws-collector-js/pull/93)

### Solution Description
Sending collector status to collector status service and it send to ingest.

- Replace the sendStatus method with sendCollectorStatus which send the status to collector status service.
- Send the streamType directly which will be set as stream while sending to collector status service

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alertlogic/paws-collector",
-  "version": "2.1.16",
+  "version": "2.1.17",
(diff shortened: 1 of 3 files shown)


imranalisyed506 (approved):
LGTM

PR #341, opened 2023-08-06
[Okta collector]: Handle the throttle error

rakhimundhada15:
### Problem Description
In integration we have seen lots of okta collectors facing throttle error which result in increasing AWS cost.

### Solution Description
Instead of returning throttle error ,now it return the same state with some poll interval delay .

- if we get the x-rate-limit-reset from header ,then add 60sec as extraBufferSeconds with reset and set the delaySecs.
- else set delaySecs by adding 60 secs with existing poll interval seconds and try adding thill max 15min delay

### DIFF
--- a/collectors/okta/okta_collector.js
+++ b/collectors/okta/okta_collector.js
@@ -19,6 +19,7 @@ const calcNextCollectionInterval = require('@alertlogic/paws-collector').calcNex
 const packageJson = require('./package.json');
 
 const THROTTLING_ERROR_REGEXP = /rateLimit/g;
+const MAX_POLL_INTERVAL = 900;
 
 const typeIdPaths = [
     { path: ['eventType'] },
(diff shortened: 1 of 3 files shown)


collectors/okta/okta_collector.js
```diff
@@ -76,8 +77,18 @@ class OktaCollector extends PawsCollector {
         .catch((error) => {
             error.errorCode = this._isNoErrorCode(error);
             if (this._isThrottlingError(error)) {
-                collector.reportApiThrottling(function() {
-                    return callback(error);
+                // if x-rate-limit-reset value return by api then accordingly delay the api call to avoid throttle error again other wise increase the delay by 1min till max 15min.
```
imranalisyed506, line 80: could you pls create a new function and add this logic there.

rakhimundhada15: move the code in _isThrottlingError

imranalisyed506 (approved):
LGTM

collectors/okta/okta_collector.js
```diff
@@ -75,9 +76,12 @@ class OktaCollector extends PawsCollector {
         })
         .catch((error) => {
             error.errorCode = this._isNoErrorCode(error);
-            if (this._isThrottlingError(error)) {
-                collector.reportApiThrottling(function() {
-                    return callback(error);
+            let isThrottle = this._isThrottlingError(error, state);
```
scott-jones-AlertLogic, line 79: just a cosmetic really, but can we rename `_isThrottlingError` and `isThrottle`? the `isX` convention is usually reserved for boolean types

maybe something like
```javascript
let maybeThrottlingError = this._resolveErrorType(error, state);
if (maybeThrottlingError.rateLimitError) {
  // ...
}
```

scott-jones-AlertLogic (commented):
thanks Rakhi, just a minor comment

scott-jones-AlertLogic (approved):
super, thank you! 🙇‍♂️

PR #344, opened 2023-08-25
[ENG-50123]: Fix the vulnerabilities related to request package for PAWS collector Lambda function 

rakhimundhada15:
### Problem Description
AWS Inspector shows latest vulnerabilities with Lambda function requires node package updates(request package) .

### Solution Description

- Replace request package with axios -https://github.com/alertlogic/al-collector-js/pull/57
- Al-aws-collector updated with latest changes of axios -https://github.com/alertlogic/al-aws-collector-js/pull/96
- Update the parameters ,header as per axios requirement .
- Return only error object received from 3rd party api and not whole axios error.

### DIFF
--- a/collectors/auth0/auth0_collector.js
+++ b/collectors/auth0/auth0_collector.js
@@ -10,7 +10,6 @@
 'use strict';
 
 const moment = require('moment');
-const request = require('request');
 const parse = require('@alertlogic/al-collector-js').Parse;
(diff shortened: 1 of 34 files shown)


imranalisyed506 (approved):
LGTM
please test all collectors which are having axios error response changes

scott-jones-AlertLogic (approved):

PR #350, opened 2023-09-22
Bump the all paws-collectors version to upgrade node version to 18

rakhimundhada15:
### Problem Description
AWS lambda runtime deprecated the node version 14

### Solution Description
Update node version to 18. 
Added aws-sdk in dependancy as node 18 runtime does not provide aws-sdk complete module.Node 18 work with aws-sdk v3 and not v2 so explicitly adding aws-sdk which is required .

### DIFF
--- a/collectors/ciscoamp/local/sam-template.yaml
+++ b/collectors/ciscoamp/local/sam-template.yaml
@@ -9,13 +9,14 @@ Resources:
       KmsKeyArn: your-arn-here
       Environment:
         Variables:
-          ssm_direct:
           AWS_LAMBDA_FUNCTION_NAME:
+          maxPages:
           aims_secret_key:
           aims_access_key_id:
           al_api:
           stack_name:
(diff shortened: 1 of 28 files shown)


imranalisyed506 (approved):

collectors/ciscoamp/local/sam-template.yaml
```diff
@@ -6,16 +6,17 @@ Resources:
   LocalLambda:
     Type: AWS::Serverless::Function
     Properties:
-      KmsKeyArn: your-arn-here
+      KmsKeyArn: arn:aws:kms:us-east-1:352283894008:key/35675069-fbc4-4954-869d-2129e0b1c145
```
scott-jones-AlertLogic, line 9: please remove key ARN from here

scott-jones-AlertLogic (approved):
super, thank you

PR #365, opened 2024-04-08
[GoogleStackDriver] Buffer portion of GCP log not being decoded for AuditLog type

imranalisyed506:
### Problem Description
[GoogleStackDriver] Buffer portion of GCP log not being decoded for AuditLog type
### Solution Description
As per google cloud [(GCPLogEntry  |  Cloud Logging  |  Google Cloud ) ](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry)
documentation the following two logs have type 
"[type.googleapis.com/google.cloud.audit.AuditLog](http://type.googleapis.com/google.cloud.audit.AuditLog)" 
send  protoPayload as buffer

image

we noticed for other customers as well when type of log as Audit “[type.googleapis.com/google.cloud.audit.AuditLog](http://type.googleapis.com/google.cloud.audit.AuditLog)" protoPayload coming as buffer.
we see all logs coming as type of log as Audit “[type.googleapis.com/google.cloud.audit.AuditLog](http://type.googleapis.com/google.cloud.audit.AuditLog)"  logs in search console for this customer, if we can get other log types for example “[type.googleapis.com/google.cloud.loadbalancing.type.LoadBalancerLogEntry](http://type.googleapis.com/google.cloud.loadbalancing.type.LoadBalancerLogEntry)” you will get to see in json format.

Note used: google-proto-files package to decode protobuf to json format for Audit log type only http://type.googleapis.com/google.cloud.audit.AuditLog).


rakhimundhada15 (approved):
Looks good

collectors/googlestackdriver/collector.js
```diff
@@ -86,7 +88,22 @@ timestamp < "${state.until}"`;
...
+                    try {
+                        const buffer = Buffer.from(logEntry.protoPayload.value);
+                        let decodedData = AuditLog.decode(buffer);
+                        logEntry.protoPayload.value = decodedData.toJSON();
+                    } catch(error) {
+                        AlLogger.error(`Error decoding data ${error}`);
+                    }
+                }
+                return logEntry;
+            });
```
scott-jones-AlertLogic, line 106: could we please move the decoding logic to a separate function/method? I would also suggest that we only load the `AuditLog` proto stuff once and then re-use it -- I think the current implementation would load it for every audit log msg we process

collectors/googlestackdriver/collector.js
```diff
@@ -86,7 +88,22 @@ timestamp < "${state.until}"`;
         const paginationCallback = (result, acc = []) => {
             AlLogger.info(`Getting page: ${pagesRetireved + 1} Logs retrieved: ${result[0].length}`);
             pagesRetireved++;
-            const logs = result[0];
+            let logs = result[0].map(logEntry => {
+                // Decode protoPayload buffer if it's a LogEntry google.cloud.audit.AuditLog
+                if (logEntry.protoPayload && (logEntry.protoPayload.type_url === 'type.googleapis.com/google.cloud.audit.AuditLog')) {
+                    const protoPath = `./node_modules/google-proto-files/google/cloud/audit/audit_log.proto`;
```
scott-jones-AlertLogic, line 94: could we pls use protoFiles.getProtoPath() here?

```node
> protoFiles.getProtoPath('cloud', 'audit', 'audit_log.proto')
'/Users/scott.jones/alertlogic/paws-collector/collectors/googlestackdriver/node_modules/google-proto-files/google/cloud/audit/audit_log.proto'
```

scott-jones-AlertLogic (commented):
thanks Imran, this looks good; please see my comment about maybe refactoring the protbuf decoding logic out of the `pawsGetLogs()` method and let me know what you think

collectors/googlestackdriver/collector.js
imranalisyed506: made changes

collectors/googlestackdriver/collector.js
```diff
@@ -16,6 +16,11 @@ const parse = require('@alertlogic/al-collector-js').Parse;
 const AlLogger = require('@alertlogic/al-aws-collector-js').Logger;
 const logging = require('@google-cloud/logging');
 const packageJson = require('./package.json');
+const protoFiles = require('google-proto-files');
+const protoPath = protoFiles.getProtoPath('cloud', 'audit', 'audit_log.proto');
+const root = protoFiles.loadSync(protoPath);
+const AuditLog = root.lookupType('google.cloud.audit.AuditLog');
```
scott-jones-AlertLogic, line 22: this feels like it belongs in a function rather than next to includes

could you please set this up in the `GooglestackdriverCollector.constructor` and make it a property of the collector obj?

collectors/googlestackdriver/collector.js
```diff
@@ -160,6 +167,19 @@ timestamp < "${state.until}"`;
             });
     }
 
+    decodeProtoPayload(logEntry) {
+        if (logEntry.protoPayload && (logEntry.protoPayload.type_url === 'type.googleapis.com/google.cloud.audit.AuditLog')) {
```
scott-jones-AlertLogic, line 171: please define the URL as a constant

scott-jones-AlertLogic (commented):
sorry for nitpicking but pls make the audit log decoder a component of the collector itself

scott-jones-AlertLogic (approved):
super; thank you, Imran! 🙇

PR #375, opened 2024-07-03
Updated the poll interval to reduce delay and handle throttle

rakhimundhada15:
### Problem Description
The customer is experiencing two delays in collecting Cisco Duo logs.

### Solution Description
Reduce the poll interval from 4 minutes (previously based on the number of streams) to 1 minute.
Change the condition to catch up if the collection delay is greater than or equal to 1 hour. [pr](https://github.com/alertlogic/paws-collector/pull/374)

According to the[ Cisco duo api doc](https://duo.com/docs/adminapi#authentication-logs)   it is recommended to make only one API call per minute. Therefore, we cannot catch up quickly, and there might be a delay of up to 1 hour, but not more than that.
image


OHaganA (approved):

rakhimundhada15:
build-collector-ciscoduo

PR #396, opened 2025-09-17
[mimecast-collector] build issue as creation of zip failed because of ESM uuid package

imranalisyed506:
### Problem Description
Upgrading uuid beyond version 10 introduces breaking changes with node version 22.3.0. Starting with uuid@12, CommonJS (require()) is no longer supported, and the package is ESM-only. Since this project still depends on CommonJS modules, using uuid@12+ results in runtime errors (ERR_REQUIRE_ESM).

### Solution Description
Pin uuid to ^10.0.0, the latest version that supports CommonJS. This ensures compatibility with our current codebase while avoiding ESM migration work for now.

### DIFF
--- a/collectors/mimecast/package.json
+++ b/collectors/mimecast/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mimecast-collector",
-  "version": "1.0.51",
+  "version": "1.0.52",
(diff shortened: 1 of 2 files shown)


imranalisyed506:
build-collector-mimecast

PR #415, opened 2026-04-01
[Cisco Duo and Cisco meraki]: Fixed the issue facing during async migration

rakhimundhada15:
### Problem Description

1. Failing self-update in Cisco Duo.
2. Sometimes the third-party API returns errors in a non-standard format (for example, a plain string instead of an object). During self-update, we add [errorCode] for DD metrics, but this fails when the error is not an object. We need to validate the error type before assigning [errorCode] so string/primitive errors are handled safely and the original error context is preserved.

### Solution Description

1. Wait and return from lambda after successful selfupdate.
2. Added a type check before assigning errorCode.
3. If the caught error is an object, we safely append ErrorCode (when missing) and rethrow.
4. If the caught error is a string/primitive, we wrap it in a standard Error object and preserve the original value.

### DIFF
--- a/collectors/ciscoduo/collector.js
+++ b/collectors/ciscoduo/collector.js
@@ -61,7 +61,7 @@ class CiscoduoCollector extends PawsCollector {
         return { state: initialStates, nextInvocationTimeout: pollInterval };
     }
 
-    pawsGetRegisterParameters(event) {
+    async pawsGetRegisterParameters(event) {
         const regValues = {
             ciscoduoObjectNames: process.env.collector_streams
         };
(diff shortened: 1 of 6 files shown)


collectors/ciscoduo/collector.js
```diff
@@ -126,8 +126,13 @@ class CiscoduoCollector extends PawsCollector {
...
             else {
-                // set errorCode if not available in error object to showcase client error on DDMetrics
-                error.errorCode = error.code;
+                // set errorCode only for object errors to avoid type errors with string/primitive throws
+                if (error && (typeof error === 'object')) {
+                    if (error.errorCode === undefined) {
+                        error.errorCode = error.code || error.status;
+                    }
+
+                }
```
scott-jones-AlertLogic: do we need to handle the case where `error` is not an object here?

scott-jones-AlertLogic (approved):

rakhimundhada15:
build-collector-ciscoduo

rakhimundhada15:
build-collector-ciscomeraki

PR #416, opened 2026-04-14
[Paws Lib]: Harden dedupe/retry flow and simplify throttling handling & Vulnerability in NPM inspected by AWS inspector

rakhimundhada15:
### Problem Description

1. In error scenarios (third-party API failure, ingest failure, or low remaining Lambda time), the collector could create retry behaviour that increased the risk of duplicate SQS state handling. 
2. CloudWatch metric throttling also generated noisy warnings during duplicate-count reporting.
3. [ENG-60148](https://helpsystems.atlassian.net/browse/ENG-60148): Vulnerability in NPM inspected by AWS inspector

### Solution Description

1. The collector retry and dedup flow was hardened to avoid duplicate state processing and reduce unnecessary retries:
- Added safer state handling to prevent duplicate in-progress processing.
- Improved retry flow to reuse progressed state and avoid replaying stale state.
- Added low-time protection so the function does not enqueue extra retry state when remaining Lambda time is too low.
- Made duplicate-count metric reporting best-effort and throttling-tolerant.
- Simplified throttling detection using common AWS error indicators

2. Updated NPM dependancy to fix the Vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alertlogic/paws-collector",
-  "version": "2.3.2",
+  "version": "2.3.3",
   "license": "MIT",
   "description": "Alert Logic AWS based API Poll Log Collector Library",
   "repository": {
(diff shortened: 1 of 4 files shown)


OHaganA (approved):

imranalisyed506 (approved):

scott-jones-AlertLogic (approved):
This looks great, thanks Rakhi! +1

PR #421, opened 2026-05-14
[All paws collectors]:Updated the packages to handle the throttle cloud watch error effectively

rakhimundhada15:
### Problem Description
These metric failures are not part of the core log collection path, but they were surfacing as noisy warnings/errors and could appear like handler instability in collectors .The result was repeated Lambda error noise every polling cycle even when collection itself was otherwise healthy

### Solution Description
Handle the cloud watch throttle error effectively in paws lib https://github.com/alertlogic/paws-collector/pull/419.
Updated the all paws collector to handle this error.

### DIFF
--- a/collectors/auth0/package.json
+++ b/collectors/auth0/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth0-collector",
-  "version": "1.2.1",
+  "version": "1.2.2",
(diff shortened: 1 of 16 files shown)


scott-jones-AlertLogic (approved):

rakhimundhada15:
build-collectors

imranalisyed506 (approved):

PR #426, opened 2026-07-16
[o365 and Cisco amp collector]Fix url.parse() deprecation warning in Node.js 24

rakhimundhada15:
### Problem Description
Node.js 24 emits [DEP0169] DeprecationWarning: url.parse() behavior is not standardized and prone to errors that have security implications on every Lambda invocation in the ciscoamp and o365 collectors

### Solution Description
Replaced url.parse() with the new URL:
- utils.js: url.parse(links.next).path → new URL(links.next) with pathname + search
- o365management.js: url.parse(requestUrl, true).query → new URL(requestUrl).searchParams.keys()
- Removed unused require('url') imports from both files

### DIFF
--- a/collectors/ciscoamp/package.json
+++ b/collectors/ciscoamp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ciscoamp-collector",
-  "version": "1.2.3",
+  "version": "1.2.4",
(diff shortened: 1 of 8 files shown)


rakhimundhada15:
build-collector-ciscoamp

rakhimundhada15:
build-collector-o365

